PR #1, opened 2021-11-19
5718 Add basic DeckGL Map with NTA layer

godfreyyeung:
### Summary
Introduces DeckGL and connects it to Carto for an NTA Overlay. A basic DeckGL map with Mapbox basemap covers the entire screen. 

#### Tasks/Bug Numbers
 - Fixes AB#5718

### DIFF
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -1,0 +1,53 @@
+import { useState } from "react";
+import { Box } from "@chakra-ui/react";
+import DeckGL from "@deck.gl/react";
+import { StaticMap } from "react-map-gl";
+import {
+  CartoLayer,
+  setDefaultCredentials,
+  API_VERSIONS,
+  MAP_TYPES,
+} from "@deck.gl/carto";
+
+setDefaultCredentials({
+  apiVersion: API_VERSIONS.V2,
+  username: process.env.NEXT_PUBLIC_CARTO_USERNAME,
+  apiKey: process.env.NEXT_PUBLIC_CARTO_API_KEY,
+});
+
+export const Map = () => {
+  const INITIAL_VIEW_STATE = {
+    longitude: -73.986607,
+    latitude: 40.691869,
+    zoom: 13,
+    pitch: 0,
+    bearing: 0,
+  };
+
+  const [layers, setLayers] = useState([
+    new CartoLayer({
+      type: MAP_TYPES.QUERY,
+      id: "dcpNta",
+      data: `SELECT * FROM dcp_nta`,
+      getLineColor: [100, 100, 100, 255],
+      getFillColor: [0, 0, 0, 1],
+      lineWidthMinPixels: 3,
+      stroked: true,
+      pickable: true,
+    }),
+  ]);
+
+  return (
+    <Box h="100%" w="100%">
+      <DeckGL
+        initialViewState={INITIAL_VIEW_STATE}
+        controller={true}
+        layers={layers}
+      >
+        <StaticMap
+          mapboxApiAccessToken={process.env.NEXT_PUBLIC_MAPBOX_TOKEN}
+        />
+      </DeckGL>
+    </Box>
+  );
+};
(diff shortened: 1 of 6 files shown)


src/pages/index.tsx
```diff
@@ -1,54 +1,12 @@
...
-  ListItem,
-} from "@chakra-ui/react";
-import { CheckCircleIcon, LinkIcon } from "@chakra-ui/icons";
-
-import { Hero } from "../components/Hero";
 import { Container } from "../components/Container";
 import { Main } from "../components/Main";
-import { Footer } from "../components/Footer";
+import { Box } from "@chakra-ui/react";
+import Map from "../components/Map/Map";
```
TylerMatteo, line 4: Once you change the component to a named export, you can create an index.ts in the `Map` folder that just has `export * from './Map'`. That will let you do `import { Map } from "../components/Map"` here, instead of needing the `/Map/Map`.

src/components/Map/Map.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const INITIAL_VIEW_STATE = {
+  longitude: -73.986607,
+  latitude: 40.691869,
+  zoom: 13,
+  pitch: 0,
+  bearing: 0,
+};
+
+export default function Map() {
```
TylerMatteo, line 26: Let's try to get out of the habit of using default exports. This can be `export const Map = () => {}`

TylerMatteo:
@godfreyyeung  Did the git hooks give you any trouble when you were committing this? I would have thought you wouldn't be able to commit if the linter is complaining.

src/components/Map/Map.tsx
```diff
@@ -0,0 +1,53 @@
...
+  MAP_TYPES,
+} from "@deck.gl/carto";
+
+setDefaultCredentials({
+  apiVersion: API_VERSIONS.V2,
+  username: process.env.NEXT_PUBLIC_CARTO_USERNAME,
+  apiKey: process.env.NEXT_PUBLIC_CARTO_API_KEY,
+});
+
+const INITIAL_VIEW_STATE = {
```
TylerMatteo, line 18: I think it would be better to move this variable declaration into the component itself. In general, it's a good practice to keep your components "pure", meaning they don't use values declared in scopes above their own.

TylerMatteo:
On second thought, I get why you were able to commit even though it failed the GH action. It's because when I reverted that one PR on the template, it also reverted a change to the git hook that included the type checking. I'd like to put that back but don't sweat it for now. Deck's typescript support is going to take some playing around with.

godfreyyeung:
> On second thought, I get why you were able to commit even though it failed the GH action. It's because when I reverted that one PR on the template, it also reverted a change to the git hook that included the type checking. I'd like to put that back but don't sweat it for now. Deck's typescript support is going to take some playing around with.

Sounds good. The husky hook did run and prevented me from committing before I fixed some of the linting issues

src/components/Map/Map.tsx
godfreyyeung: Sounds good!

godfreyyeung: I've always been neutral and not sure if one is better than the other

src/components/Map/Map.tsx
godfreyyeung: Hmm okay will change that. Curious to learn why

src/components/Map/Map.tsx
godfreyyeung: Hmm, but actually in this case `INITIAL_VEW_STATE` is acting more like a private variable. If the intention is that it is never modified or publicly accessible, doesn't it make sense to keep "privatize" it by placing it in the module scope?

TylerMatteo: It is effectively "private" if it's declared inside the function. To clarify, I'm suggesting still defining it as a const just like it is here (as opposed to as some component state). I think declaring it outside the function scope would actually make it more "global". Keeping thing inside the scope of the function also makes writing tests easier to write - one less value you have to mock.

TylerMatteo (approved):

PR #4, opened 2021-11-29
Add new /map page with optional NTA code param

godfreyyeung:


src/pages/map/[[...nta]].tsx
```diff
@@ -0,0 +1,126 @@
...
+      onClick: (info: any) => {
+        const id = info && info?.object?.properties?.id;
+        if (id && typeof id === "string" && id !== selectedNta) {
+          router.push(`/map/${id}`, undefined, { shallow: true });
+        }
+      },
+      // For performance reasons, Deck doesn't automatically rerender when external data changes
+      // updateTriggers lets us re-evaluate a given property when a given value changes
+      // https://deck.gl/docs/developer-guide/performance#use-updatetriggers
+      updateTriggers: {
```
Comment (godfreyyeung) [inline src/pages/map/[[...nta]].tsx:87]: clever -- vs adding a separate layer just for fill altogether

godfreyyeung:
With this approach we'll need to figure out redirecting the user to `/map` on application load

src/pages/map/[[...nta]].tsx
```diff
@@ -0,0 +1,126 @@
...
+  // }, []);
+
+  // useEffect(() => {
+  //   const { nta } = router.query;
+  //   const newSelectedNta: string | null =
+  //     nta && nta?.length > 0 ? nta[0] : null;
+  //   setSelectedNta(newSelectedNta);
+  // }, [router.query?.nta[0]]);
+
+  // I changed this to be a regular variables instead of state because the event handler wouldn't work
```
Comment (godfreyyeung) [inline src/pages/map/[[...nta]].tsx:57]: interesting

src/pages/map/[[...nta]].tsx
Comment (TylerMatteo) [reply src/pages/map/[[...nta]].tsx:87]: Yea this is maybe an example of where the terminology isn't one to one between mapbox style spec and deck. Within Deck, the geospatial representations of the NTAs are one layer, their visual representation is just an implementation detail that is derived from the data.

src/pages/map/[[...nta]].tsx
Comment (TylerMatteo) [reply src/pages/map/[[...nta]].tsx:57]: Yea I wasn't planning on changing this until it started causing issues because I could imagine us wanting to have the layers array in state eventually, even if we don't need it to be for now. Still, a good rule of them is that something probably doesn't need to be state if you never call its setter function.

TylerMatteo:
> With this approach we'll need to figure out redirecting the user to `/map` on application load

Yea good call, I'll go ahead and implement that in this branch.

src/pages/nta/[[...nta]].tsx
```diff
@@ -0,0 +1,126 @@
...
+      <DeckGL
+        initialViewState={INITIAL_VIEW_STATE}
+        controller={true}
+        layers={layers}
+      >
+        <StaticMap
+          mapboxApiAccessToken={process.env.NEXT_PUBLIC_MAPBOX_TOKEN}
+          mapStyle={baseMap}
+        />
+        <VStack
```
Comment (godfreyyeung) [inline src/pages/nta/[[...nta]].tsx:104]: I think this floating box should probably be outside of the Map component?

next.config.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  async redirects() {
+    return [
+      {
+        source: "/",
```
godfreyyeung, line 5: cool!

src/pages/nta/[[...nta]].tsx
```diff
@@ -0,0 +1,126 @@
...
+  // const [selectedNta, setSelectedNta] = useState<string | null>(null);
+
+  // useEffect(() => {
+  //   const { nta } = router.query;
+  //   const initialSelectedNta: string | null =
+  //     nta && nta?.length > 0 ? nta[0] : null;
+  //   setSelectedNta(initialSelectedNta);
+  // }, []);
+
+  // useEffect(() => {
```
Comment (godfreyyeung) [inline src/pages/nta/[[...nta]].tsx:50]: Was this an accidental duplicate comment?

TylerMatteo (approved):

PR #5, opened 2021-12-07
5718 Adding mock nta indicator data and basic responsive map page layout

TylerMatteo:
### Summary
This PR adds an extremely basic page structure and some content interaction to the existing `/nta` map page, loosely based on the low fidelity figma wireframes. It adds a json file of static nta-level indicator data. Then, it uses that data to show indicators for an nta when a user selects one from the map. It also colors in the ntas based on the `displacementRisk` value in the mock data.

Please note that the project at this point **does not** necessarily represent a best-practices way of creating reusable components or how to implement a design system with Chakra. Components here such as `<Map>` are not intended to be generic and are intentionally purpose built for this project.

#### Tasks/Bug Numbers
 - Fixes AB#5718


src/hooks/useSelectedNta.ts
```diff
@@ -0,0 +1,20 @@
+import { useRouter } from "next/router";
+import ntas from "@data/ntas.json";
+import { NtaIndicatorRecord } from "../types/Nta";
+
```
TylerMatteo, line 4: In case it's not totally obvious what's going on here - this is a custom react hook to return the mock data corresponding to the nta id currently reflected in the URL. Pulling in the Next.js router context, doing some null-checking, and matching the id to the mock data takes a handful of lines of code that were cropping up in a couple places so I decided to encapsulate that code here.

src/pages/nta/[[...nta]].tsx
```diff
@@ -1,33 +1,20 @@
...
-          >
-            Clear Selection
-          </Button>
-        </VStack>
-      </Map>
+    <Box height="100vh">
+      <Header />
+      <Map />
+      {!isMobile && (
+        <Legend position="absolute" left={8} bottom={8} w="215px" />
```
Comment (TylerMatteo) [inline src/pages/nta/[[...nta]].tsx:15]: You'll probably notice that this `<Legend>` component gets conditionally rendered twice - once here for non-mobile screen sizes, and again inside the `<IndicatorPanel>` component for mobile screen sizes. That's not ideal but haven't quite figured out a better way to implement it. It's tricky because on mobile, the Legend wants to be "inside" the same panel as the indicators because it needs to derive its height based on it being there, where as desktop the legend is a totally separate thing rendered in the bottom left corner.

godfreyyeung (changes requested):
Hey this looks awesome, but it would be great if you could  break this out into multiple PRs. This looks like at least four things smooshed into one commit in one PR. It'll be a lot easier to review when it's broken out into focused changes. I think it's also really important for us to get into the practice of making focused commits and PRs as we talked about in our Engineering Practices discussion 

There looks to be separate additions here for tests, UI structure, NTA data querying, map styling, Legend work, data ingestion, etc.

src/pages/nta/[[...nta]].tsx
Comment (TylerMatteo) [reply src/pages/nta/[[...nta]].tsx:15]: Ok actually disregard this, I figured out a better way to handle this

TylerMatteo:
Gotcha. I'll see if I can roll back my commit and break this into a couple PRs - one with just blank panels of the UI, and then another to add the mock data and querying stuff.

TylerMatteo:
Closing this PR in favor of forthcoming PRs that break this work up.

PR #6, opened 2021-12-08
5718 Basic layout for map page

TylerMatteo:
### Summary
This PR adds a basic UI to the [[nta]] (map) page including a header, legend, and panel for showing the indicators for the selected NTA.

#### Tasks/Bug Numbers
 - Fixes AB#5718


src/components/IndicatorPanel/IndicatorPanel.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export const IndicatorPanel = () => {
+  const { isOpen, onToggle } = useDisclosure();
+
+  return (
+    <Box
+      background="#fff"
+      position={["relative", "absolute"]}
+      right={["auto", "0px"]}
+      top={["auto", "68px"]}
```
godfreyyeung, line 12: just curious -- what is auto and the 68px here indicating? The hstack above is has `top=4.25rem`. should this match that?

src/components/Map/Map.tsx
```diff
@@ -30,47 +26,43 @@ export const Map = ({ children = null }: MapProps) => {
...
-          : [0, 0, 0, 1],
+      getFillColor: [0, 0, 0, 0],
       lineWidthMinPixels: 3,
       stroked: true,
       pickable: true,
       onClick: (info: any) => {
-        const id = info && info?.object?.properties?.id;
-        if (id && typeof id === "string" && id !== selectedNta) {
+        const id: any =
+          info && info?.object?.properties?.id
```
godfreyyeung, line 43: just wondering - is it redundant to use both `info &&` and `info?.` ? Doesn't `?.` already check `info` is not undefined or null?

godfreyyeung (dismissed):
Thank you for breaking it up a bit! Looks good to me. Nit: the deckgl typings could still be moved to another PR

src/components/IndicatorPanel/IndicatorPanel.tsx
TylerMatteo: The items in that array correspond to what the style should be at each breakpoint, starting with the smallest. The first is just auto because the panel isn't absolutely positioned on small screens. As for the 68px, it should be "4.25rem" so I'll change that. Those sizes are equivalent for users who don't enlarge the text in their browser but we should be doing rems across the board for consistency.

src/components/Map/Map.tsx
TylerMatteo: I double checked how optional chaining works and it looks like it is redundant actually. Some trivia - it looks like the only time you would need the `info &&` is if you were worried about `info` not being initialized at all but of course your IDE is going to yell at you anyway if you're using variables that don't exist.

src/components/IndicatorPanel/IndicatorPanel.tsx
TylerMatteo: Also, this is probably a good moment to call out that once we have a design system and Chakra theme more fleshed out, we probably wouldn't be setting values like `height`,`top`, or `margin` with rems or px explicitly at all, but instead with regular old integers or strings from our Theme's design tokens.

godfreyyeung: > Also, this is probably a good moment to call out that once we have a design system and Chakra theme more fleshed out, we probably wouldn't be setting values like `height`,`top`, or `margin` with rems or px explicitly at all, but instead with regular old integers or strings from our Theme's design tokens.

sounds good. Yeah  I was thinking  these inline styles would start being an anti-pattern...

TylerMatteo: Yea it's something I need to get around to explaining how it's supposed to work, versus how we're handling it to get this project off the ground. It's helpful to think of styles applied like this as not exactly "inline" in the traditional sense because Chakra actually compiles these styles to a unique generated class and puts that class name on the div, just like you would normally do by hand. This approach just saves you from having to come up with classnames and switch between js and css files. Still, you're correct that having a ton of these "style props" on one component is often a sign that you should move them to some sort of reusable variant or component.

godfreyyeung: Ah, thanks for explaining that! That's good to know!

godfreyyeung (approved):

PR #12, opened 2022-01-06
Feature/5992 ci improvements

TylerMatteo:
### Summary
This PR implements a suite of Github Actions to handle code quality and test checks for the project before finally doing preview, staging, and production deployments manually via the Netlify CLI. To facilitate this, I also added a stubbed-out `getServerSideProps` function the the main map page. This is because Next automatically determines whether or not it should statically export the entire site or run a node server to do server side generation based on whether or not any pages implement SSR. Whether or not we have SSR also effects the workflow for building and publishing the site so, even though we don't actually need to implement SSR to do anything for this project in it's current state, I stubbed it out because it's likely we'll need it in the future.

The `netlify.toml` file was updated to add the [official netlify plugin for nextjs](https://github.com/netlify/netlify-plugin-nextjs). This plugin is necessary because, by default, netlify only deploys static sites whereas Next.js projects can do server side rendering and therefore require a running node process. This plugin takes the server-side functionality of our site and automatically converts them to be handled by Netlify functions under the hood.

There are 3 github action workflows:
* `feature_branch.yml` runs when there is a push to a branch matching the glob pattern `feature/[0-9]+-*` (ex: feature/93848-some-description). It will lint and typecheck the code before simultaneously kicking off a preview deploy to Netlify, e2e tests, and unit tests
* `pr_merged.yml` runs when there is a push to `main` that isn't a semver version tag. This is intended to run when, and only when, a PR into `main` is merged. It will lint and type check, then wait for unit and e2e tests to pass, then deploy to a netlify staging url
* `release_and_deploy.yml` is triggered manually from the actions page for this repo on GitHub. This means we can merge into main without necessarily kicking off a release to production. This workflow is associated with the protected "production" environment in the repo so when someone kicks it off manually, it will wait until an approver approves the workflow before actually running. Once that happens, it will checkout HEAD of `main`, lint, and test. Assuming that succeeds, the action will run `npm version` as the `planner-bot` Github user. `npm version` does a few things - first it will increment the version in `package.json` by a "patch" increment (this is hardcoded for now but we should add functionality for major and minor soon). Then it will commit the changed package.json file to main (as `planner-bot`, which is the only account that should be allowed to commit to main), then add a git tag to the new commit that matches the version. Finally, it will push the new commit and tag to `main` and kick off a production deployment to Netlify via the CLI.

#### Tasks/Bug Numbers
 - Fixes AB#5992


TylerMatteo:
Closing this because I made a fresh branch to cover the same work -> https://github.com/NYCPlanning/equity-tool/pull/15

PR #13, opened 2022-01-06
6154 5986 Replay re-architecture work onto main

godfreyyeung:
Replays re-architecture work onto the main branch. 

App is now focused around a map catch-all subpage, and uses hooks to acquire state from URL segments.

Fixes AB#6154


src/components/IndicatorPanel/IndicatorPanel.tsx
```diff
@@ -1,9 +1,9 @@
 import { Box, Collapse, useDisclosure, Text, Button } from "@chakra-ui/react";
 import { ChevronUpIcon, ChevronDownIcon } from "@chakra-ui/icons";
-import { useSelectedNta } from "@hooks/useSelectedNta";
 
-export const IndicatorPanel = () => {
-  const selectedNta = useSelectedNta();
+export const IndicatorPanel = (props) => {
```
TylerMatteo, line 4: Two things here
-  it's a common pattern in React to destructure out of `props` directly where you're declaring the function
- You'll need to define a typescript interface that describes the the type for `props`, so something like

```
interface IndicatorPanelProps {
  indicatorRecord: NtaIndicatorRecord
}

export const IndicatorPanel = ({ indicatorRecord }: IndicatorPanelProps) => {
...
```

src/components/Map/GeographySelect/GeographySelect.tsx
```diff
@@ -0,0 +1,32 @@
+import { useRouter } from "next/router";
+import { Button, ButtonGroup } from "@chakra-ui/react";
+
+// Docs on BoxProps?
+export const GeographySelect = (props) => {
```
TylerMatteo, line 5: Same deal as comment on IndicatorPanel

src/components/Map/Map.tsx
```diff
@@ -1,27 +1,18 @@
...
-import { useSelectedNta } from "@hooks/useSelectedNta";
 
 setDefaultCredentials({
   apiVersion: API_VERSIONS.V2,
   username: process.env.NEXT_PUBLIC_CARTO_USERNAME,
   apiKey: process.env.NEXT_PUBLIC_CARTO_API_KEY,
 });
 
-export const Map = () => {
+export const Map = (props) => {
```
TylerMatteo, line 13: Aaaand again, sorry I know this is annoying to not catch until it's pushed to Github. These errors would be caught by the commit hook if lint-staged was doing type checking so I'll try to fix that

src/components/IndicatorPanel/IndicatorPanel.tsx
godfreyyeung: Okay will do that!

TylerMatteo (approved):

PR #23, opened 2022-01-19
6307 Basic Data (Table) route and page stub. Share Flex container and Header with map page

godfreyyeung:
Creates the basic /data route, and stubs the Data Page.

Centralizes the containing Flex  and Header components, so they are shared across main pages.

image

Fixes AB#6307


src/pages/map/[[...subroutes]].tsx
```diff
@@ -50,7 +49,8 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
     /map/geography/geoid
 */
 const MapPage = ({ initialRouteParams }: MapPageProps) => {
-  console.log(initialRouteParams);
+  console.log(initialRouteParams); // only here to prevent unused variable initialRouteParams?
```
Comment (TylerMatteo) [inline src/pages/map/[[...subroutes]].tsx:52]: yea :(

src/pages/data/[geoid].tsx
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  return {
+    props: {
+      initialRouteParams: subroutes ? subroutes.join(",") : "",
+    },
+  };
+};
+
+const Sidebar = () => {
```
Comment (TylerMatteo) [inline src/pages/data/[geoid].tsx:34]: Are we sure this sidebar is only used on the data page? I vaguely remember it showing up on the map page when we looking at the mobile prototype. If it does, we should probably move this to its own file in `src/components`. If it doesn't, I'm cool with how this is set up.

src/pages/_app.tsx
```diff
@@ -6,7 +9,13 @@ import { AppProps } from "next/app";
 function MyApp({ Component, pageProps }: AppProps) {
   return (
     <ChakraProvider resetCSS theme={theme}>
-      <Component {...pageProps} />
+      <Flex height="100vh" direction="column" bg="gray.100">
```
TylerMatteo, line 12: Do we really need the two `<Flex>`'s here? It feel like we could keep `_app` more generic but just rending `<Component>` inside a full width/height `<Box>` or something.

src/pages/data/[geoid].tsx
Comment (godfreyyeung) [reply src/pages/data/[geoid].tsx:34]: Definitely it will be it's own component file eventually, but will leave that for another dedicated PR. Hmm so it is meant to represent this blue bar below. I've only seen this on the data page so far 
image

src/pages/_app.tsx
godfreyyeung: The outer Flex here helps the container on line 15 fill the rest of the screen below the Header. 
If there's no flex, the elements inside Box don't know how to properly stretch to just fill the space underneath the header. Or maybe i'm overlooking a simple solution. Any ideas for a substitute to this approach?

TylerMatteo: I just thought maybe we could get away with just a `<Box>` here but I can see how this is functionally the same, or better. I'm good with this solution

src/pages/data/[geoid].tsx
Comment (TylerMatteo) [reply src/pages/data/[geoid].tsx:34]: Gotcha, I'm cool with leaving it here for the purposes of this PR. We can move it out as necessary.

TylerMatteo (approved):
lgtm

PR #24, opened 2022-01-20
6308 Quick, crude stub of About page

godfreyyeung:
Fixes AB#6308

### DIFF
--- a/src/pages/about.tsx
+++ b/src/pages/about.tsx
@@ -1,0 +1,31 @@
+import { Flex, Box, Center } from "@chakra-ui/react";
+
+const AboutPage = () => {
+  return (
+    <Flex direction="column" width="100%">
+      <Flex direction="row" bg="teal" height="300px">
+        <Box flex="1"></Box>
+        <Box flex="5">
+          <h1>About</h1>
+        </Box>
+        <Box flex="2">
+          <h2>Feedback</h2>
+        </Box>
+        <Box flex="1"></Box>
+      </Flex>
+      <Box>
+        <Center width="100%">
+          <Flex direction="row">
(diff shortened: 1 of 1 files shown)


src/pages/about.tsx
```diff
@@ -0,0 +1,31 @@
+import { Flex, Box, Center } from "@chakra-ui/react";
+
+const AboutPage = () => {
+  return (
+    <Flex direction="column" width="100%">
+      <Flex direction="row" bg="teal" height="300px">
+        <Box flex="1"></Box>
+        <Box flex="5">
+          <h1>About</h1>
```
TylerMatteo, line 9: Not worried about this for a stub, just making a note we can get into the habit of using [Chakra's heading component](https://chakra-ui.com/docs/typography/heading) for these.

src/pages/about.tsx
```diff
@@ -0,0 +1,31 @@
+import { Flex, Box, Center } from "@chakra-ui/react";
+
+const AboutPage = () => {
+  return (
+    <Flex direction="column" width="100%">
+      <Flex direction="row" bg="teal" height="300px">
+        <Box flex="1"></Box>
```
TylerMatteo, line 7: I think "spacer divs" like these are an anti-pattern. What we probably want here is a centered container with a max-width.

src/pages/about.tsx
godfreyyeung: Okay will try that out

src/pages/about.tsx
godfreyyeung: Yeah, just stubbing for now so made it really quick.

TylerMatteo (approved):

PR #25, opened 2022-01-24
6338 Use MapContext to render NavigationControls

godfreyyeung:
Fixes AB#6338
See  https://deck.gl/docs/api-reference/react/deckgl#react-context

image

### DIFF
--- a/src/components/Map/Map.tsx
+++ b/src/components/Map/Map.tsx
@@ -7,6 +7,8 @@ import ReactMapGL, {
 import { setDefaultCredentials, API_VERSIONS } from "@deck.gl/carto";
 import baseMap from "@data/basemap.json";
 
+import { Box } from "@chakra-ui/react";
+
 setDefaultCredentials({
   apiVersion: API_VERSIONS.V2,
   username: process.env.NEXT_PUBLIC_CARTO_USERNAME,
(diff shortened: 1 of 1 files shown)


src/components/Map/Map.tsx
```diff
@@ -21,17 +24,30 @@ export const Map = ({ layers, parent }: MapProps) => {
...
   const map = (
     <DeckGL
       initialViewState={INITIAL_VIEW_STATE}
       controller={true}
       layers={layers}
       parent={parent}
+      ContextProvider={MapContext.Provider}
     >
-      <StaticMap
+      <div style={navControlStyle}>
```
TylerMatteo, line 43: Could you replace this with some style props on a `<Box>`? Or do your styles get overridden when you do it that way?

TylerMatteo: As another option, according to the docs it looks like you can pass `style` directly to `NavigationControl` so you don't need the extra div.

godfreyyeung: I tried both (Box and directly to NavigationControl), and that doesn't work. passing it to navigationcontrol directly was the first thing I tried and couldn't get that to work for the longest time. Maybe I'm missing something

src/components/Map/Map.tsx
```diff
@@ -21,17 +24,30 @@ export const Map = ({ layers, parent }: MapProps) => {
...
   };
 
+  const navControlStyle = {
+    position: "absolute",
+    bottom: 150,
+    left: 10,
+  } as React.CSSProperties;
+
+  // MapContext is necessary for navigation controls to work.
+  // Unclear on why.
```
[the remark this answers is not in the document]
godfreyyeung: Yeah haha that's the same link I put in the PR description

src/components/Map/Map.tsx
TylerMatteo: Hmm that's odd. I pulled your branch and changed it to just `<NavigationControl style={navControlStyle} />` and it seems to work, I just had to play with the value for "bottom" to get it to match up with what you have.

godfreyyeung: Okay good call. It wasn't working when NavigationControl was nested under ReactMapGL, but it's okay when it's outside of it

TylerMatteo (approved):

PR #29, opened 2022-01-26
6389 ViewSelect (DRI/Data Tool) desktop toggle 

godfreyyeung:
Fixes AB#6389

This implements the toggle to switch between DRI and Data Tool routes. It preserves the selected geography and geoid in either view. We assume DRI only operates in PUMA so far.

image


godfreyyeung:
Hold for tests

godfreyyeung:
Ready for review

src/hooks/useSelectedLayer/useSelectedLayer.ts
```diff
@@ -7,75 +7,80 @@ import { interpolateRgb } from "d3-interpolate";
...
 export const useSelectedLayer = (
+  view: string | null,
   geography: string | null
 ): CartoLayer<any, any>[] | null => {
   const router = useRouter();
 
   const scale = scaleSequential().domain([0, 100]);
   const interpolate = interpolateRgb("#f4f4b4", "#d44932");
 
-  switch (geography) {
```
TylerMatteo, line 17: Not something to worry about for this PR but just making a note that we should eventually refactor this to be less bloated considering most of the CartoLayer properties are the same between geo types.

src/components/Map/ViewSelect.tsx
```diff
@@ -0,0 +1,29 @@
...
+    <ButtonGroup isAttached {...boxProps}>
+      <Button
+        onClick={onDataToolClick}
+        isActive={view === "datatool"}
+        data-cy="dataToolBtn"
+      >
+        Data Tool
+      </Button>
+      <Button onClick={onDriClick} isActive={view === "dri"} data-cy="driBtn">
+        DRI
```
TylerMatteo, line 25: It looks like "DRI" is written out in Figma. Also, I think we should avoid adding data props just for cypress to have something to select. It ends up recreating the same problem that Chakra solves by removing the need to come up with (and keep track of) class names made up be devs. 

That being said, it looks like [I forgot to set up Cypress to have access to React Testing Library's query API](https://testing-library.com/docs/cypress-testing-library/intro/). With that, you should be able to use the React Testing Library API inside of cypress tests and select things from the DOM with `getByText`, `getByRole`, etc. - instead of adding stuff to our React code that exists just to facilitate tests.

If you want to take a crack at adding that RTL integration for Cypress, I think it would be great to handle now so that we don't end up bloating our markup with data props but I wouldn't hold up this PR over it.

TylerMatteo (changes requested):
I added some comments on things to address but the only one I would hold up merging over is the styling.

src/components/Map/ViewSelect.tsx
godfreyyeung: Hmm, I was going off of Cypress the best practices guide: https://docs.cypress.io/guides/references/best-practices#Selecting-Elements

src/components/Map/ViewSelect.tsx
```diff
@@ -0,0 +1,29 @@
+import { BoxProps, Button, ButtonGroup } from "@chakra-ui/react";
+
+type ViewSelectProps = {
```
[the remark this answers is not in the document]
godfreyyeung: yep!

src/components/Map/ViewSelect.tsx
```diff
@@ -0,0 +1,29 @@
...
+} & BoxProps;
+
+export const ViewSelect = ({
+  onDataToolClick,
+  onDriClick,
+  view,
+  ...boxProps
+}: ViewSelectProps) => {
+  return (
+    <ButtonGroup isAttached {...boxProps}>
```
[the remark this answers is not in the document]
godfreyyeung: What part of the styling? We should wait until later to make it look exactly like the Figma in terms of color and shadow and degree of rounded-ness, etc. Let's focus on functionality and user stories first

src/hooks/useSelectedLayer/useSelectedLayer.ts
godfreyyeung: Sounds good to me

src/hooks/useSelectedLayer/useSelectedLayer.ts
```diff
@@ -7,75 +7,80 @@ import { interpolateRgb } from "d3-interpolate";
 import ntas from "@data/ntas.json";
 
 export const useSelectedLayer = (
+  view: string | null,
```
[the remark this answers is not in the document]
TylerMatteo: There are also [actual enums](https://www.typescriptlang.org/docs/handbook/enums.html#enums-at-compile-time) which are useful for when you need to be able to access the actual list of potential values are runtime. I totally agree that we wouldn't want to necessitate checking the value before it's passed in. In a lot of scenarios, a literal type would actually save you from having to do that because the TS compiler will yell at you if you try to pass in an invalid string but I see how that doesn't work for our case here because the value you're passing in comes off the router so we can't make that sort of assumption about it. Fine with this as is.

src/components/Map/ViewSelect.tsx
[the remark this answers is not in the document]
godfreyyeung: I think we probably want to set up a styling system that's a bit more formal and applied consistently throughout the app, so there should be a separate, dedicated and focused effort & discussion to create the styles exactly to spec and apply it to all components. For example creating variables for the fonts, hover styles, background color, rounded-ness, etc, and then making sure they apply to all relevant components. I'm wary of spinning too much on it in this early phase of the build, and think it will be easier once all the core functionality and DOM element settles down. We could do it at this point, but at least it should be a separate PR so the discussion for this one doesn't balloon. It'll involve setting up variables common to other components

src/components/Map/ViewSelect.tsx
TylerMatteo: Hmph, that is interesting. I'm cool with this as is, sounds like there's more of a discussion to be had here. I'm more used to using RTL for DOM selection in tests, and I think adding it to cypress can be nice because then we're using the same "api" in unit and e2e tests. That being said, it looks like [their guidance](https://testing-library.com/docs/guiding-principles) is at odds with the Cypress best practices. Oh well, for now we can carry on writing our Cypress tests according to Cypress' guidance.

godfreyyeung: Yeah I hear you about using the same RTL API, that does sound nice. In this case I had to move away from querying by text because it was brittle -- the tests broke because the Drawer also had the text Data Tool.  
My current opinion is that using specific test selector makes the test faster to read and debug in the future. Maybe that will change after using the RTL api a bit

TylerMatteo: Yea that all makes sense to me. Another possible way we could choose to look at it could be as simple as "Prefer RTL for unit (jest) tests, and prefer cypress' built-ins for e2e". Not something we have to come to an agreement on now, I'm just happy to have tests tbh

src/components/Map/ViewSelect.tsx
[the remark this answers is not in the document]
TylerMatteo: Hmm yea I hear you about wanting to fill out the tokens a bit more, I'm just concerned about spending the time on that before April. It looks like Design has been using tokens from the default Chakra theme (teal.500, gray.50, etc) in Figma and I've been using those token in my PRs but getting too invested into those defaults has its own set of risks. 

I just don't want us to burn too much time defining those sorts of things given our deadlines. Styling the built-in components [the Chakra way](https://chakra-ui.com/docs/theming/component-style#styling-single-part-components), including pseudos, variants, etc, is pretty straight forward but gets a little confusing when you start having to define your own "custom" components like this switcher thing. I'm happy to talk irl about what would be a good baseline of things to define now and how that should translate to code.

TylerMatteo (approved):

PR #31, opened 2022-01-27
Task/6309 Mobile Drawer

godfreyyeung:
Fixes [AB#6309](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/6309) 

This implements the mobile Drawer (the desktop sidebar turns into a bottom drawer when screen size shrinks)

image

### DIFF
--- a/src/components/Map/MobileDrawer.tsx
+++ b/src/components/Map/MobileDrawer.tsx
@@ -8,7 +8,7 @@ interface MobileDrawerProps {
 }
 
 export const MobileDrawer = ({ title, children }: MobileDrawerProps) => {
-  const [isOpen, setIsOpen] = useState(true);
+  const [isOpen, setIsOpen] = useState(false);
 
   return (
     <Box
@@ -25,6 +25,7 @@ export const MobileDrawer = ({ title, children }: MobileDrawerProps) => {
       left="0"
       zIndex="999"
       bg="white"
+      data-cy="mobileDrawer"
     >
       <Flex direction="column" height="100%">
         <Flex
@@ -52,6 +53,7 @@ export const MobileDrawer = ({ title, children }: MobileDrawerProps) => {
                 onClick={() => {
                   setIsOpen(!isOpen);
                 }}
+                data-cy="closeMobileDrawer"
               />
             ) : (
               <IconButton
@@ -59,6 +61,7 @@ export const MobileDrawer = ({ title, children }: MobileDrawerProps) => {
                 aria-label="Toggle Drawer"
                 fontSize="20px"
                 icon={<ChevronUpIcon />}
+                data-cy="openMobileDrawer"
               />
             )}
           </Box>
--- a/src/components/Map/MobileDrawer.tsx
+++ b/src/components/Map/MobileDrawer.tsx
@@ -1,0 +1,73 @@
+import { useState } from "react";
+import { Heading, Box, IconButton, Flex } from "@chakra-ui/react";
+import { ChevronUpIcon, ChevronDownIcon } from "@chakra-ui/icons";
+
+interface MobileDrawerProps {
+  title: string | null;
+  children?: React.ReactNode;
+}
+
+export const MobileDrawer = ({ title, children }: MobileDrawerProps) => {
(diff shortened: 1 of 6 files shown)


godfreyyeung:
Needs some tests

godfreyyeung:
Ready for review

src/pages/map/[[...subroutes]].tsx
```diff
@@ -5,6 +5,7 @@ import { Box } from "@chakra-ui/react";
 import { useSelectedLayer } from "@hooks/useSelectedLayer";
 import { useIndicatorRecord } from "@hooks/useIndicatorRecord";
 import { IndicatorPanel } from "@components/IndicatorPanel";
+import { MobileDrawer } from "@components/Map";
 import { Map } from "@components/Map";
```
Comment (TylerMatteo) [inline src/pages/map/[[...subroutes]].tsx:9]: You can import MobileDrawer and Map in one import statement if they both fall under `Map`

src/components/Map/MobileDrawer.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { Heading, Box, IconButton, Flex } from "@chakra-ui/react";
+import { ChevronUpIcon, ChevronDownIcon } from "@chakra-ui/icons";
+
+interface MobileDrawerProps {
+  title: string | null;
+  children?: React.ReactNode;
+}
+
+export const MobileDrawer = ({ title, children }: MobileDrawerProps) => {
+  const [isOpen, setIsOpen] = useState(false);
```
TylerMatteo, line 11: I'm cool with just using `useState` but just fyi Chakra has a little utility hook for this sort of thing called [useDisclosure](https://chakra-ui.com/docs/hooks/use-disclosure)

src/components/Map/MobileDrawer.tsx
```diff
@@ -0,0 +1,78 @@
+import { useState } from "react";
+import { Heading, Box, IconButton, Flex } from "@chakra-ui/react";
+import { ChevronUpIcon, ChevronDownIcon } from "@chakra-ui/icons";
+
+interface MobileDrawerProps {
+  title: string | null;
+  children?: React.ReactNode;
+}
+
+export const MobileDrawer = ({ title, children }: MobileDrawerProps) => {
```
TylerMatteo, line 10: Do you think you'll end up renaming this to something like just `Drawer` when you do the desktop work? I was thinking we'd implement this as one responsive component but I'm curious if you've run into issues what could make that impractical.

TylerMatteo:
If you can fix that redundant `import` real quick I'm cool with approving this. I was going to ask if you tried using Chakra's [collapse](https://chakra-ui.com/docs/components/transitions#collapse-transition) component instead of managing the height yourself but I noticed it actually doesn't support expanding horizontally so that solution would break down anyway when we go to do desktop. (there's a [feature request](https://github.com/chakra-ui/chakra-ui/issues/2791) about it but sounds like it's not a priority for them).

src/components/Map/MobileDrawer.tsx
godfreyyeung: Oo, thanks. Will try that next time. I tend to prefer simplifying to "primitives" when possible to avoid library specific APIs, so future devs can understand immediately. But I also see the convenience of that hook.

src/components/Map/MobileDrawer.tsx
godfreyyeung: I guess in the future we could wrap this MobileDrawer and the DesktopSidebar into one component....
Though currently we need to display completely separate Boxes for the desktop sidebar and mobile drawers.
And seems conceptually/design wise one is a sidebar and another is a drawer, so it might not make sense to just have a Drawer component

TylerMatteo (approved):

src/components/Map/MobileDrawer.tsx
TylerMatteo: Yeah you can make that call when you sit down to do the desktop work. I thought that they were both always at least partially visible and then had expanded "states". And (in theory) we could handle the differences, like anchoring to the left vs anchoring to the bottom, with responsive styling

PR #32, opened 2022-01-28
6403 Added about page content

dhochbaum-dcp:
- Completes [AB#6403](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/6403)

This adds all of the text and image content to the "About" page.


TylerMatteo:
Do you see this weirdness when you pull this up in a mobile screen size?
image

Also, it looks like there's some inconsistency with the background color on "shorter" screen sizes
image

(I think the background below the teal section should be all white)

TylerMatteo:
Oh yea, looks like that first issue goes away if you delete the header. I'll see if it's fixed when you merge my header stuff with your changes here so don't worry about that one for now. For the gray background thing, please just delete that `bg="gray.100"` in _app.tsx - the most up to date figmas don't have the padding around the individual elements of the Map page so it's no longer necessary and removing that will fix this.

src/components/About/CustomDivider.tsx
```diff
@@ -0,0 +1,11 @@
+import { Center, Divider } from "@chakra-ui/react";
+
+const CustomDivider = () => {
```
TylerMatteo, line 3: Does this really warrant a separate component? You should be able to add margin or padding directly to the instance of `<Divider>`

TylerMatteo (approved):

PR #34, opened 2022-02-02
6389 Style view toggle

godfreyyeung:
Finalizes [AB#6389](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/6389)

Styles the toggle according to specs through the Chakra UI default theme.

Safari:
image
image

Chrome (browser introduces blue highlights around links). TODO: figure out how to apply webkit styles in chakra ui to prevent the highlighting
image
image

### DIFF
--- a/src/theme.tsx
+++ b/src/theme.tsx
@@ -16,6 +16,38 @@ const theme = extendTheme({
   },
   fonts,
   breakpoints,
+  components: {
+    ButtonGroup: {
+      parts: ["group", "button"],
+      variants: {
+        toggle: {
+          group: {
+            backgroundColor: "white",
+            borderRadius: 50,
+          },
+        },
+      },
+    },
+    Button: {
+      variants: {
+        aqua: {
+          backgroundColor: "white",
+          color: "gray.600",
+          borderRadius: 50,
+
+          _hover: {
+            color: "teal",
+          },
+          _active: {
+            backgroundColor: "teal.50",
+            border: "1px solid teal",
+            color: "teal",
+            borderRadius: 50,
+          },
+        },
+      },
+    },
+  },
 });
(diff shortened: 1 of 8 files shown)


src/theme.tsx
```diff
@@ -16,6 +16,38 @@ const theme = extendTheme({
...
+          group: {
+            backgroundColor: "white",
+            borderRadius: 50,
+          },
+        },
+      },
+    },
+    Button: {
+      variants: {
+        aqua: {
```
TylerMatteo: I'll leave this up to you, and we can revisit names of things later when we start defining components more clearly with Design, but my two cents is to favor "semantic" variant names over ones that just describe the style.

godfreyyeung: Yah actually in a subsequent PR (Geography Toggle PR) i renamed it to "toggle"

godfreyyeung:
> That blue outline is coming from some `:focus` styling applying by chakra's default and serves a purpose for accessibility. I'm cool with changing it but we do need to make sure there is a way for users who rely on the keyboard for interacting with the site to know what element has focus. If you load the page and `tab` key to the unselected option, it currently looks like this: image
> 
> If you get rid of the that focus style, there would be no way for the user to know where they are. One easy solution that would probably work is just to make sure the style applied to `_hover` is also applied to `_focus`. If the user can see the font color change, that's probably enough of a queue that we can safely get rid of the outline.

Gotcha. I forgot about that, good catch. I'd rather leave it then actually, to preserve Chrome's default accessibility behavior.

TylerMatteo (dismissed):

TylerMatteo (approved):

PR #36, opened 2022-02-02
6571 Style GeographySelect toolbar

godfreyyeung:
Fixes [AB#6571](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/6571) 

Safari:
image

### DIFF
--- a/src/theme.tsx
+++ b/src/theme.tsx
@@ -16,6 +16,38 @@ const theme = extendTheme({
   },
   fonts,
   breakpoints,
+  components: {
+    ButtonGroup: {
+      parts: ["group", "button"],
+      variants: {
+        toggle: {
+          group: {
+            backgroundColor: "white",
+            borderRadius: 50,
+          },
+        },
+      },
+    },
+    Button: {
+      variants: {
+        aqua: {
+          backgroundColor: "white",
+          color: "gray.600",
+          borderRadius: 50,
+
+          _hover: {
+            color: "teal",
+          },
+          _active: {
+            backgroundColor: "teal.50",
+            border: "1px solid teal",
+            color: "teal",
+            borderRadius: 50,
+          },
+        },
+      },
+    },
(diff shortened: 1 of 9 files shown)


src/theme.tsx
```diff
@@ -16,6 +16,56 @@ const theme = extendTheme({
...
+          color: "teal",
+        },
+        _active: {
+          backgroundColor: "teal.50",
+          border: "1px solid teal",
+          color: "teal",
+        },
+      },
+      variants: {
+        leftCap: {
```
TylerMatteo, line 46: It's probably not worth getting into this now but to me this feels like a misuse variants. If you look at the source code for `<ButtonGroup>`, it's setting those border radii with regular css `first-of-type` pseudo selectors which feels more idiomatic. If you have to apply a particular variant to each button within a button group for it to look how you want, doesn't that defeat the purpose?

src/theme.tsx
```diff
@@ -16,6 +16,56 @@ const theme = extendTheme({
...
+          borderRadius: "50px 0 0 50px",
+          borderRight: "1px solid #A0AEC0",
+        },
+        middle: {
+          borderRadius: 0,
+          borderRight: "1px solid #A0AEC0",
+          _active: {
+            marginLeft: "-1px", // prevents double border
```
TylerMatteo, line 54: Are you sure you need the border that this is meant to account for at all? In the [hand off](https://www.figma.com/file/U7dgahwMaggEmMI1l5O1Z9/Equitable-Development-Reporting---EDDT-Tool?node-id=2378%3A47598) it looks like these buttons just get a background and font color change when they're active or hovered.

godfreyyeung: This is what I saw: 
image

TylerMatteo (approved):

PR #38, opened 2022-02-03
6587 Implement styled Mobile View Toggle

godfreyyeung:
Fixes [AB#6587](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/6587)
image

### DIFF
--- a/src/theme.tsx
+++ b/src/theme.tsx
@@ -45,19 +45,6 @@ const theme = extendTheme({
         },
       },
       variants: {
-        bigToggle: {
-          backgroundColor: "gray.50",
-          height: "3.75rem",
-          _first: {
-            borderRadius: ".5rem 0 0 0",
-          },
-          _last: {
-            borderRadius: "0 .5rem 0 0",
-          },
-          _active: {
-            border: "1px solid teal",
-          },
-        },
         leftCap: {
           borderRadius: "50px 0 0 50px",
           borderRight: "1px solid #A0AEC0",
--- a/src/components/Map/ViewToggle.tsx
+++ b/src/components/Map/ViewToggle.tsx
@@ -22,13 +22,22 @@ export const ViewToggle = ({
         direction="row"
         position="fixed"
         width="100%"
+        height="3.75rem"
         bottom="0"
         left="0"
         zIndex="999"
       >
         <Button
           flex="1"
-          variant="bigToggle"
+          variant="toggle"
+          backgroundColor="gray.50"
+          height="100%"
+          borderRadius=".5rem 0 0 0"
+          _active={{
+            backgroundColor: "teal.50",
+            color: "teal",
+            border: "1px solid teal",
+          }}
           onClick={onDataToolClick}
           isActive={view === "datatool"}
           isFullWidth
@@ -38,7 +47,15 @@ export const ViewToggle = ({
         </Button>
         <Button
           flex="1"
-          variant="bigToggle"
+          variant="toggle"
+          backgroundColor="gray.50"
+          height="100%"
+          borderRadius="0 .5rem 0 0"
+          _active={{
+            backgroundColor: "teal.50",
+            color: "teal",
+            border: "1px solid teal",
+          }}
           onClick={onDriClick}
           isActive={view === "dri"}
           isFullWidth


cypress/integration/pages/Map/[[...subroutes]].spec.ts
```diff
@@ -32,33 +32,33 @@ describe("Map catch-all page", () => {
     it("should switch view when user uses ViewToggle toolbar", () => {
       cy.url().should("include", "/map/datatool");
 
-      cy.get('[data-cy="driBtn"]').click();
+      cy.get('[data-cy="driBtn"][data-cy-context="desktop"]').click();
```
Comment (godfreyyeung) [inline cypress/integration/pages/Map/[[...subroutes]].spec.ts:35]: i'm iffy about this, but for some reason cypress.get() doesn't recognize attributes with non-string values (like `is-mobile={true}` )

Comment (godfreyyeung) [reply cypress/integration/pages/Map/[[...subroutes]].spec.ts:35]: Here i needed to add this extra attribute, because there are two buttons with `data-cy="driBtn"`. One for mobile, one for desktop.

src/theme.tsx
```diff
@@ -45,6 +45,19 @@ const theme = extendTheme({
         },
       },
       variants: {
+        bigToggle: {
```
TylerMatteo, line 48: Are we really sure we need to render separate buttons for mobile vs desktop? I'm not convinced this can't be done with responsive styling. It feels like we're getting very far away from what is supposed to constitute a "variant". If something like this is a variant, it's going to get out of hand pretty quickly.

cypress/integration/pages/Map/[[...subroutes]].spec.ts
Comment (TylerMatteo) [reply cypress/integration/pages/Map/[[...subroutes]].spec.ts:35]: I think the fewer cypress data attributes we have, the better. If we absolutely must have separate mobile and desktop buttons, I'd rather just do something like `data-cy="driBtn-mobile` and `data-cy="driBtn-desktop`. Every time one of us makes up an attribute name like this, it establishes another pseudo-convention that other devs have to keep in their head.

src/theme.tsx
godfreyyeung: If we do this through responsive styling, would all Buttons of "toggle" variant jump between small and big buttons? Seems awkward and a way too specific of a variant. The buttons are different enough that it's better to capture the logic in the templates, rather than in styles, imo.

cypress/integration/pages/Map/[[...subroutes]].spec.ts
Comment (godfreyyeung) [reply cypress/integration/pages/Map/[[...subroutes]].spec.ts:35]: Sure. Yeah, at this point it's about agreeing on a convention.

src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -1,11 +1,16 @@
...
   useMultiStyleConfig,
 } from "@chakra-ui/react";
 
+interface ToggleButtonGroupInterface extends BoxProps {
+  isAttached?: boolean;
+}
+
 // In this case we want implicit children
-export const ToggleButtonGroup: React.FC<{ isAttached?: boolean }> = (
+export const ToggleButtonGroup: React.FC<ToggleButtonGroupInterface> = (
```
TylerMatteo, line 13: I don't mind this for now but eventually we should decide whether or not we're going to use these interfaces that ship with React (such as `FC`). I don't have the links handy but I've read that the best convention going forward is to not to use them and just type `children` as `React.ReactNode` when you have a component with children.

cypress/integration/pages/Map/[[...subroutes]].spec.ts
Comment (godfreyyeung) [reply cypress/integration/pages/Map/[[...subroutes]].spec.ts:35]: > And I think the problem with `{true}` stuff is because Cypress is testing against the fully rendered app running in a real browser, so it's reading the actual html. I'd imagine something like `<Box data-cy-is-mobile="true">` and `cy.get('[data-cy-is-mobile="true"]')` might work

ah. right, but at that point it looks kinda gross to me to have `true` in strings lol, so decided against it

src/components/ToggleButtonGroup/ToggleButtonGroup.tsx
godfreyyeung: Hmm yeah, I saw that React.FC is "frowned upon" but it seemed it was due to the implicit children. I'm not knowledgeable to have an opinion on typing `children` as  `React.ReactNode`

src/theme.tsx
TylerMatteo: If "toggle" applies only to the "binary" switcher we're using for DRI vs Data tool, then yes they would all behave that way. I don't love that solution either but honestly these should all be totally separate components that are composed of buttons. I'm just trying to limit tech debt at this point. Going from small to big based on screen size is exactly the kind of thing that should be done by styles, whenever possible, imo (granted those styles could be in-line or via variants)

TylerMatteo: If the responsive styles here (going between full width and auto, between rounded corners or not, etc) really are specific to this particular pair of buttons, then this is strictly a one-off and shouldn't pollute our variant names

PR #42, opened 2022-02-07
6665 Sidebar/Drawer Data Tool "Welcome" content

godfreyyeung:
This PR primarily inserts the "Welcome" content for the Data Tool view into the sidebar and mobile drawer.
It improves the clarity of the DOM architecture within the Sidebar/mobile drawer. 

A subsequent PR for [AB#6719](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/6719) needs to conditionally switch the Sidebar and Drawer based on the view selected (Data Tool or DRI)  

This PR also
  -  makes sure that the ViewToggle (is hidden and) does not obscure the Mobile Drawer when a geography is selected 
  - Repurposes the StaticFooter for the sidebar/drawer. Cleans up the architecture surrounding the StaticFooter on the About and Methods pages. 

Fixes [AB#6665](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/6665)

image

image

image

image


src/pages/map/[[...subroutes]].tsx
```diff
@@ -117,23 +185,52 @@ const MapPage = ({ initialRouteParams }: MapPageProps) => {
...
     <>
-      <Box
+      <Flex
         display={{
           base: "none",
           lg: "flex",
         }}
+        direction="column"
+        justify="space-between"
         flex="1"
```
Comment (TylerMatteo) [inline src/pages/map/[[...subroutes]].tsx:195]: This is sort of unrelated to this PR but just wanted to note that this sidebar should have a[ fixed width,](https://www.figma.com/file/U7dgahwMaggEmMI1l5O1Z9/Equitable-Development-Reporting---EDDT-Tool?node-id=3366%3A90349) not flex.

src/components/Map/ViewToggle.tsx
```diff
@@ -5,65 +5,69 @@ interface ViewToggleProps extends BoxProps {
...
-          flex="1"
-          variant="toggle"
-          backgroundColor="gray.50"
-          height="100%"
-          borderRadius=".5rem 0 0 0"
-          _active={{
-            backgroundColor: "teal.50",
-            color: "teal",
-            border: "1px solid teal",
+      {showToggle && (
```
TylerMatteo, line 19: Can you explain why we need this? When would ever want to hide the view toggle component?

TylerMatteo: If we do need it, I generally think adding props that effectively show or hide a whole component is an anti-pattern in React. In that scenario you should just conditionally render the whole component in the parent component where you're using it.

src/components/Map/MobileDrawer.tsx
```diff
@@ -20,8 +20,8 @@ export const MobileDrawer = ({ title, children }: MobileDrawerProps) => {
       width="100%"
       position="fixed"
       top={isOpen ? "6rem" : "100vh"}
-      marginTop={isOpen ? "auto" : "-15rem"}
-      paddingBottom="12.5rem"
+      marginTop={isOpen ? "auto" : "-7.875rem"}
```
TylerMatteo, line 23: We should revisit how to build this drawer without negative margins. Ideally we wouldn't be implementing a "collapsed" state by pushing content off the screen. Negative margins always lead to odd bugs down the road in my experience, and they can complicate testing writing and accessibility.

src/components/Map/ViewToggle.tsx
godfreyyeung: The spec is that the ViewToggle is hidden when a geography is selected. Confirmed with Natasha
image

src/components/Map/MobileDrawer.tsx
godfreyyeung: Okay, let's create another task for this

src/components/Map/ViewToggle.tsx
godfreyyeung: In this case, we can't conditionally render the whole component because only a part of the component (mobile) is hidden when a geography is selected. Happy to break this into two components, though (MobileViewToggle and DesktopViewToggle)

src/pages/map/[[...subroutes]].tsx
Comment (godfreyyeung) [reply src/pages/map/[[...subroutes]].tsx:195]: Sounds good, lets file a bug

src/pages/methods.tsx
```diff
@@ -195,7 +195,11 @@ const MethodsPage = () => {
         </Flex>
       </Center>
 
-      <StaticPageFooter />
```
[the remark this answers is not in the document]
godfreyyeung: In this case, the ` <Center><Box width={["100%", "1024px"]} py="2rem">` is actually a pattern specific to this file `methods.tsx`. Each segment of the page uses this pattern to wrap and contain/center its content. 
I think later we can revisit and wrap the entire page in something like [Container](https://chakra-ui.com/docs/layout/container) FYI @dhochbaum-dcp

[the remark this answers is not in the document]
godfreyyeung: Added 😄

TylerMatteo (approved):

PR #47, opened 2022-02-14
Task/6653 Position navigation control according to spec

godfreyyeung:
Fixes AB#6653

- Positiones the nav control according to figma spec. 
- Desktop: sits top right, under the View Toggle
- Mobile: sits top right, at corner edge if in DRI view. Otherwise, just under the Geography Toggle in Data Tool view.

### DIFF
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -9,7 +9,7 @@ import { AppProps } from "next/app";
 function MyApp({ Component, pageProps }: AppProps) {
   return (
     <ChakraProvider resetCSS theme={theme}>
-      <Flex height="100vh" direction="column">
+      <Flex height="100vh" direction="column" overflow="hidden">
         <Header />
 
         <Flex direction="row" flex="auto">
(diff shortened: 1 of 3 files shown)


src/pages/_app.tsx
```diff
@@ -9,7 +9,7 @@ import { AppProps } from "next/app";
 function MyApp({ Component, pageProps }: AppProps) {
   return (
     <ChakraProvider resetCSS theme={theme}>
-      <Flex height="100vh" direction="column">
+      <Flex height="100vh" direction="column" overflow="hidden">
```
TylerMatteo, line 12: What necessitates this? I don't think we should be putting any more styles on these divs in `_app` because it powers all of the pages. Could this be on the outer-most component of a particular page?

godfreyyeung: It could be applied to a child container, but it acts as a catch-all for flickering issues involving overflow. It's a single page application and this is the outer wrapper of the SPA, so it should only help. I'm okay with removing it though if you have strong opinions about it.

TylerMatteo: If it's to fix that flickering issue, I'm cool with it just living here. I'm a little concerned that it could cause issues when we implement things like the tables that have horizontal scrolling but we can deal with that if/when it becomes an issue. Although, it's worth pointing out that this isn't technically a SPA because we're using Next. That being said, I can see how we're basically treating the map stuff as a little SPA on it's own and I don't mind putting things in `_app` until it starts to bork other pages.

TylerMatteo (approved):

PR #49, opened 2022-02-14
Task/6803 Style and format geo information box for Map View - Data Tool 

godfreyyeung:
Fixes  [AB#6803](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/6803) 

image

Formats and styles geo info box according to Figma spec

### DIFF
--- a/src/components/SidebarContent/SidebarContent.tsx
+++ b/src/components/SidebarContent/SidebarContent.tsx
@@ -1,5 +1,6 @@
 import React from "react";
-import { Box, Heading } from "@chakra-ui/react";
+import { Heading, Box, Button } from "@chakra-ui/react";
+import { CloseIcon } from "@chakra-ui/icons";
 import WelcomeContent from "@components/WelcomeContent";
 import WelcomeFooter from "@components/WelcomeFooter";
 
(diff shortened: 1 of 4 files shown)


src/components/SidebarContent/SidebarContent.tsx
```diff
@@ -12,18 +13,29 @@ export const SidebarContent = ({
...
     return (
-      <Box>
-        <Heading>PUMA 132</Heading>
-        <br />
-        Either Data Tool Categories or DRI Indicators go here <br />
-        (you can use useView() to determine which)
-      </Box>
+      <>
+        <Box paddingBottom="2rem">
+          <Heading fontSize=".8125rem" fontWeight={500} color="teal.600">
```
TylerMatteo, line 18: Maybe a little nit-picky here but you ought to have h1-h6 show up in the DOM in numerical order (I believe Chakra defaults `<Heading>` to an h2 if you don't pass an `as` prop). I was playing around with an implementation of this in my Data Explorer branch where I put these three headings in a column flex in their "semantic" order and then used the flex prop `order` to get them to show up in the order they are in the design. I took it out because it was out of scope for that PR but it seemed to work pretty well. Idk how picky we want to be about semantic/accessibilty stuff like this atm but just something to consider.

TylerMatteo (approved):

src/components/SidebarContent/SidebarContent.tsx
godfreyyeung: Ah didn't know about the order attribute. Yeah design wise the geography id is definitely a subheader, and the geography label is header, so was trying to code it semantically to that. Interesting, I haven't heard of how the h1-h6 elements need to be in order in the DOM first. If a user tabs twice, I assume the first tab would select the geography label "woodside & sunnyside", then the second one is the geography id. Which seems like the correct order to me?

TylerMatteo: The rule of thumb I try to follow is in they should show up in order, and yea, ideally not skipping any. This case is a little ambiguous and one could argue the puma id, neighborhood label, and "approx cd" are all one big heading anyway. I just came cross the `order` thing as a way to handle the (very rare) scenario when you want stuff to show up on the screen in non-semantic order, so thought I'd share :D. Screen-readers will ignore `order` and read things in the order they appear in the DOM.

PR #54, opened 2022-02-16
6855 Connect GeographySelect btns to correct Carto layers

godfreyyeung:
- Fixes [AB#6855](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/6855)

Stepping stone PR for work on the Data Tool sidebar. 

We are going to use the `dcp_puma` layer  for Community District because it is so called "Data Pipeline"-ed (maintained by GIS). Although it doesn't have neighborhood information, Erica has given us a lookup table for the PUMA description and CD associations per PUMA.

### DIFF
--- a/src/hooks/useSelectedLayer/useSelectedLayer.ts
+++ b/src/hooks/useSelectedLayer/useSelectedLayer.ts
@@ -1,10 +1,5 @@
 import { useRouter } from "next/router";
 import { CartoLayer, MAP_TYPES } from "@deck.gl/carto";
-import { scaleSequential } from "d3-scale";
-import { rgb } from "d3-color";
-import { interpolateRgb } from "d3-interpolate";
-
-import ntas from "@data/ntas.json";
 
 export const useSelectedLayer = (
   view: string | null,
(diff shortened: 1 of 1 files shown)


src/hooks/useSelectedLayer/useSelectedLayer.ts
```diff
@@ -1,41 +1,33 @@
...
   if (view === "datatool") {
     switch (geography) {
       case "district":
         return [
           new CartoLayer({
             type: MAP_TYPES.QUERY,
-            id: "censusarea",
-            data: `SELECT * FROM pff_2020_census_tracts_21c`,
+            id: "puma",
+            data: `SELECT * FROM dcp_puma`, // todo: update to correct carto dataset
```
TylerMatteo, line 17: Just wondering - are you choosing not to hook this up to the dataset we will really be using because you want to keep the scope of the PR small or because you weren't sure which one to use? I only ask because I was also having trouble figuring out which one we should be using for PUMAs. There are two datasets `dcp_puma_2010` and `dcp_puma` that seem to be well maintained but lack the information and labels about each puma we need. On the other hand there is `nyc_puma` which seems to have all the data we need but might be less maintained and doesn't specify the year of the data.

I'm cool with revisiting this source in another PR but if we aren't clear on what source we should be using, I'd like to flag that.

TylerMatteo (dismissed):

src/hooks/useSelectedLayer/useSelectedLayer.ts
godfreyyeung: @TylerMatteo  Oops i should have deleted that comment. I decided that `dcp_puma` should be the right one because it is the official maintained one. It doesn't have the neighborhood description values but we can get that from a static file.

TylerMatteo: Ok works for me. I guess I pictured that stuff coming from carto and the static file only having the indicator data plus a geoid to match on but I don't mind doing it this way. I guess my dream of having "source of truth" data like that live in Carto to be used across apps is probably too much to ask for anyway haha

TylerMatteo (approved):

PR #55, opened 2022-02-17
6778 DataTable and EstimateTable components (mobile)

TylerMatteo:
### Summary
This PR begins to build out common components that can eventually be used across the different categories within the Data Explorer. Note that the styling work here is really only meant to meet the mobile specs but my intention is that these components will be expanded upon or tweaked to accommodate desktop as well

Breakdown of what's included:
* Some changes in `theme.tsx` that tweak the already existing `striped` variant of Chakra's `<Table>` component to match our design specs
* A new `DataTable` (with `DataTableHead` and `DataTableBody` sub-components). This component is meant to encapsulate the common behavior found in all tables as seen on mobile. That includes the horizontal scrolling, ability to hide the table by clicking on the header, and any styling that is outside the scope of the `striped` variant. 
* A new `EstimateTable` component that uses DataTable to build a table for data points that are "estimates", which, in our case corresponds to the data points that come from PUMS and have moe, cv, etc. It handles the logic for showing/hiding reliability metrics. Note that it also handles the styling for the "sticky" first column. I went back and forth on putting that here vs building it directly in DataTable, since it's likely that all of our tables will have that functionality but I couldn't settle on a clean way of doing it that would account for the subtle differences in tables. I'm definitely open to feedback here and we can visit how these components fit together as we build out more of the app.
* Some simple types for "Estimates". These exist to show the structure I'm imagining we'll have for the data that fits this "Estimate" concept. These may also change as we build out more of the app but these felt like a good starting point.
* Some placeholder data and markup in `[geoid].tsx`. This is meant to make the reliability metric toggling testable and roughly show how multiple `DataTables` will stack horizontally on larger screens.

Things not in this PR that will be coming soon
* Styling for screen sizes >= `md`. I added some placeholder data and basic styling to show how multiple `DataTables` can stack horizontally on larger screen sizes but obviously there's a lot more styling to do there. I'll probably eventually have some sort of `DataTableGroup` component to handle some of that.
* The ability to hide or show all the tables on a given page. I'll probably need to add some props to `DataTable` so that it has a way to trigger this.
* Other components that are specific to a particular "shape" of data in the same way that `EstimateTable` is. We can have ones for data that is not an estimate and therefore can be built with the assumption that it doesn't have cv, moe. I'll also be adding a component and types for the Change Over Time tables.

#### Tasks/Bug Numbers
 - Fixes [AB#6778](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/6778)


src/components/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,84 @@
...
+  if (!context) throw new Error("DataTableContext Provider not found");
+  return context as DataTableContextReturn;
+};
+
+export const DataTable = ({ children }: DataTableProps) => {
+  const { isOpen, onToggle } = useDisclosure({
+    defaultIsOpen: true,
+  });
+  // Table is always expanded if on size md or larger, else is expanded if expanded flag is true
+  const isMobile = useWindowWidth() < 768;
```
godfreyyeung, line 29: Just curious, is there no way to reference a centralized variable/constant for this width? Seems odd to have static breakpoint values  littered throughout

src/components/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  return (
+    <DataTableContext.Provider
+      value={{ isOpen: isOpen || !isMobile, onToggle }}
+    >
+      <Box marginLeft={{ base: 3, md: 0 }} overflowX={"scroll"}>
```
godfreyyeung, line 35: prefer specifying units ... `3rem`?

src/components/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,84 @@
...
+            whiteSpace: "nowrap",
+            "thead tr": {
+              th: {
+                minW: "calc((100vw - 26px) / 3)",
+                maxW: "calc((100vw - 26px) / 3)",
+                whiteSpace: "normal",
+              },
+              "&:first-of-type": {
+                "th:first-of-type": {
+                  borderTopLeftRadius: "12px",
```
godfreyyeung, line 54: Thought we prefer `rem`?

src/components/DataTable/DataTableHeader.tsx
```diff
@@ -0,0 +1,58 @@
...
+            <Box
+              display={{ base: "block", md: "none" }}
+              position={"absolute"}
+              left={"1.5rem"}
+            >
+              {isOpen ? (
+                <ChevronDownIcon
+                  color="teal.600"
+                  _hover={{ color: "teal.600" }}
+                  w={10}
```
godfreyyeung, line 40: prefer units

src/components/DataTable/DataTable.tsx
TylerMatteo: I think we can go ahead with using the unitless values when it's being done intentionally. This value is supposed to evaluate to `0.75rem` so here I'm using the existing size token.

src/components/DataTable/DataTable.tsx
TylerMatteo: Actually, yes there is. I should be able to grab the breakpoint out of the theme with `useTheme`. I'll make that change.

src/components/DataTable/DataTableHeader.tsx
TylerMatteo: See above comment

godfreyyeung (approved):

PR #57, opened 2022-02-17
6877 6883 Data Tool Selected Geography Information

godfreyyeung:
Fixes [AB#6877](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/6877) [AB#6883](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/6883)

### DIFF
--- a/src/components/SidebarContent/DataTool/GeographyInfo.tsx
+++ b/src/components/SidebarContent/DataTool/GeographyInfo.tsx
@@ -1,0 +1,47 @@
+import { CloseIcon } from "@chakra-ui/icons";
+import { Heading, Box, Button } from "@chakra-ui/react";
+import { useMapSubrouteInfo } from "@hooks/useMapSubrouteInfo";
+import { usePumaInfo } from "@hooks/usePumaInfo";
+
+const getGeographyLabel = (geographyId: string | null): string | null => {
+  switch (geographyId) {
+    case "borough":
+      return "Borough";
+    case "district":
+      return "Community District";
+    case "citywide":
+      return "City";
+  }
+
+  return null;
+};
+
+export const DataToolGeographyInfo = () => {
+  const { geography, geoid } = useMapSubrouteInfo();
+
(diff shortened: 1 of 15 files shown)


src/components/SidebarContent/DataTool/GeographyInfo.tsx
```diff
@@ -0,0 +1,50 @@
+import { CloseIcon } from "@chakra-ui/icons";
+import { Heading, Box, Button } from "@chakra-ui/react";
+import { useMapSubrouteInfo } from "@hooks/useMapSubrouteInfo";
+import { usePumaInfo } from "@hooks/usePumaInfo";
+
+export const DataToolGeographyInfo = () => {
```
TylerMatteo, line 6: A couple things about this component:
* What do you think about having separate components for geography type? That way you can conditionally render the correct one instead of having all this logic in the component. 
* Can you move this component to just be in the top level "components" folder? I'll eventually be reusing it on data explorer. In general, I don't really like this pattern of namespacing stuff in the components folder - it feels like it makes things hard to find and is being influenced by some heavy-handed ember conventions.

src/components/SidebarContent/DataTool/GeographyInfo.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const { geography, geoid } = useMapSubrouteInfo();
+
+  const pumaInfo = usePumaInfo(geoid);
+
+  let primaryHeading = "";
+
+  switch (geography) {
+    case "district":
```
TylerMatteo, line 14: Can you make an enum of constants for these values? I think it's useful to have a `/src/constants` folder for stuff like this.

godfreyyeung:
Borked by conflicts. Closing in favor of #59 #60  and...

PR #61, opened 2022-02-23
Task/6942 citywide geoid

godfreyyeung:
Fixes AB#6942


src/constants/geoid.ts
```diff
@@ -0,0 +1,3 @@
+export enum Geoid {
+  Nyc = "nyc",
```
TylerMatteo, line 2: Do you imagine this ever having more than just `Nyc`? If not, I think this might be a little confusing since we treat most geoids as a piece of data that could be effectively "any" value.

TylerMatteo (approved):

PR #63, opened 2022-02-23
Task/6893 explorer category selection

TylerMatteo:
### Summary
This PR adds a new `CategoryMenu` component to the Data Explorer page. This responsive component is meant to be eventually reused on the map page, albeit maybe with some additional changes to it. I also added a set of constants for the categories themselves. The Data Explorer layout will be slightly off with these changes because of work that needs done to make `DataTable` play nice on `md` screen sizes but that will be in another PR.

#### Tasks/Bug Numbers
 - Fixes [AB#6893](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/6893)


src/components/CategoryMenu/CategoryMenu.tsx
```diff
@@ -0,0 +1,68 @@
...
+} from "@fortawesome/free-solid-svg-icons";
+
+export interface CategoryMenuProps extends FlexProps {
+  currentCategory?: Category;
+  geography: Geography;
+  geoid: string;
+}
+
+export const CategoryMenu = ({
+  geography,
```
godfreyyeung, line 20: Can you just acquire `geography` and `geoid` from `useMapSubrouteInfo()`?

src/constants/Category.ts
```diff
@@ -0,0 +1,7 @@
+export enum Category {
+  DEMO = "demo",
```
godfreyyeung, line 2: What do you think about doing CamelCase for enums, and all caps for consts?

src/components/CategoryMenu/CategoryMenu.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { BuildingHouseIcon, MentalHealthIcon } from "@components/Icons";
+import {
+  faUserGroup,
+  faUmbrella,
+  faHouseUser,
+} from "@fortawesome/free-solid-svg-icons";
+
+export interface CategoryMenuProps extends FlexProps {
+  currentCategory?: Category;
```
godfreyyeung, line 14: can you just grab the category from the URL?

src/components/CategoryMenu/CategoryMenu.tsx
godfreyyeung: And/or update useMapSubrouteInfo to return geography of type Geography  enum

src/constants/Category.ts
TylerMatteo: I like CamelCase for enums but I think the individuals members of the enum (as well as stand-alone constants) should be all caps. My reasoning is that in TS, CamelCase usually refers to types (with the unfortunately exception of react components) and enums are essentially types themselves but the individual members of them are not.

src/components/CategoryMenu/CategoryMenu.tsx
TylerMatteo: This component is used in the data explorer page (`/data/[geography]/[goeid]/[category].tsx`) so no, that hook wouldn't work for me here. I built this component to take geography, geoid, and currentCategory as props so that the component isn't coupled to either page. You can use `useMapSubrouteInfo()` to grab the data to pass in on the data page just like I use `useDataExplorerState()` to grab the data here.

src/components/CategoryMenu/CategoryMenu.tsx
TylerMatteo: See reply to first comment

src/constants/Category.ts
TylerMatteo: Take this with a huge grain of salt because we don't claim to adhere to any particular open source style guide for TS but fwiw [Google's TS guide does all caps for enum members ](https://google.github.io/styleguide/tsguide.html#identifiers). But, again, big'ol grain of salt, you could definitely find guides saying otherwise without looking too hard.

godfreyyeung: Okay, that's fine. Was asking just to make sure we're consistent. The enums in `constants/Geography` are CamelCased, so we should change those. Do you want to make that quick change? I'm happy to do it, just let me know.

godfreyyeung (approved):

godfreyyeung (approved):

PR #65, opened 2022-02-24
6941 Capture view and geography as explicit dynamic segments, geoid as queryParam

godfreyyeung:
- Creates specific dynamic segments for `view` and `geography`
- Updates `useMapSubrouteInfo`() to pull those segments from router.query, instead of router.query.subroutes
- Updates useMapSubrouteInfo to make use of Geography enums, and with more streamlined type checking
- Updates transitions to push `geoid` as queryParameter. 
- Switching away from using the `pathname: {}` signature for router.push because it encodes `?=` and breaks the URL. 
- Removes the redundant hook useView(), which was only causing errors due to mismatch with new dynamic segment architecture.

Fixes [AB#6941](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/6941)


dhochbaum-dcp (dismissed):
Changes look good but unfortunately they break the homepage, please fix.

godfreyyeung:
> Changes look good but unfortunately they break the homepage, please fix.

Can you be more specific? Are you sure you aren't loading a cached version?

TylerMatteo (approved):

PR #68, opened 2022-02-25
Task/6969 data explorer geo info

TylerMatteo:
### Summary
This PR refactors the `<GeographyInfo>` component so that it can be reused on the Data Explorer page, and then adds it there. In order to decouple that component from the `[geography].tsx` map page, I changed the component to get its data via props instead of directly calling `useMapSubrouteInfo`. I also moved the "clear selection" button out of that component because those don't exist on the DE page.

#### Tasks/Bug Numbers
 - Fixes [AB#6969](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/6969)


src/components/GeographyInfo/GeographyInfo.tsx
```diff
@@ -35,42 +36,26 @@ export const GeographyInfo = () => {
   }
 
   return (
-    <Box paddingBottom="2rem" flex="shrink" title={primaryHeading}>
-      {view === "datatool" && geography === DISTRICT && (
-        <Heading fontSize=".8125rem" fontWeight={500} color="teal.600">
+    <Box flex="shrink" title={primaryHeading}>
+      {geography === DISTRICT && (
+        <Heading fontSize=".8125rem" fontWeight={500} color="#2B797A">
```
godfreyyeung, line 41: why change from teal.600 to the hex value?

src/components/SidebarContent/SidebarContent.tsx
```diff
@@ -1,19 +1,40 @@
...
-        <GeographyInfo />
-        <hr />
+        <GeographyInfo
+          geoid={geoid}
+          geography={geography}
+          fontSize="1.5625rem"
+          isTruncated
+        />
+        {geoid !== NYC && (
+          <Button
```
godfreyyeung, line 26: maybe we should make this a ClearGeographyButton component

src/components/SidebarContent/SidebarContent.tsx
```diff
@@ -1,19 +1,40 @@
...
+            rightIcon={<CloseIcon />}
+            variant="outline"
+            size="xs"
+            onClick={clearSelection}
+            alignSelf={"start"}
+          >
+            Clear Selection
+          </Button>
+        )}
+        <Divider color={"gray.200"} my={"1.5rem"} />
```
godfreyyeung, line 37: If we're replacing `<hr/>` with this, there's a few more that should be replaced as well. In DrawerContent and in DRISelection

godfreyyeung: Should probably make it a component, too

src/components/GeographyInfo/GeographyInfo.tsx
TylerMatteo: Design had us change this color because teal.600 doesn't pass color contrast on top of the gray DE sidebar but this alternative they gave doesn't have a name in the Design System :/

src/components/SidebarContent/SidebarContent.tsx
TylerMatteo: I would be down for that but I'm pretty sure we're going to be removing this button entirely with the "card" updates anyway so maybe it's ok to leave it like this in this PR?

src/components/SidebarContent/SidebarContent.tsx
TylerMatteo: Hmm Idk if it warrants another abstraction on top of what `<Divider>` already does, considering the margin is different from one usage to the next. I was also considering adding bottom borders to some of the <Box> or <Flex> components instead of using Divider so much, but stayed away from that here because I wanted to change Map page stuff as little as possible in this PR. What would that new component look like?

src/components/GeographyInfo/GeographyInfo.tsx
godfreyyeung: ahh gotcha

src/components/SidebarContent/SidebarContent.tsx
godfreyyeung: Basically just one line, ` <Divider color={"gray.200"} my={"1.5rem"} />`, so they all have the same color and my haha. Np

godfreyyeung (approved):

PR #69, opened 2022-02-26
6954 Acquire NTA Geo Info from carto layer

godfreyyeung:
Acquires selected geography information from Carto, and renders it in the sidebar/drawer.
image

Another approach is to pull this information from ntas.json... but there are some discrepancies in the ntacode values 

[AB#6954](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/6954)

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -9,6 +9,6 @@ module.exports = {
     ];
   },
   env: {
-    ntaLayer: "dcp_nta_2010",
+    NTA_LAYER: "dcp_nta_2010",
   },
 };
--- a/next.config.js
+++ b/next.config.js
@@ -8,4 +8,7 @@ module.exports = {
       },
     ];
   },
+  env: {
+    ntaLayer: "dcp_nta_2010",
+  },
 };
--- a/src/helpers/fetchNtaInfo.ts
+++ b/src/helpers/fetchNtaInfo.ts
@@ -1,7 +1,7 @@
 export const fetchNtaInfo = async (
   ntacode: string | null,
   onNtaInfo: (ntaInfo: any) => void
-): Promise<any> => {
+): Promise<null> => {
   if (!ntacode) return null;
 
   const ntaInfoSql = `SELECT ntacode, ntaname FROM ${process.env.ntaLayer} WHERE ntacode='${ntacode}'`;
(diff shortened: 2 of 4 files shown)


next.config.js
```diff
@@ -8,4 +8,7 @@ module.exports = {
       },
     ];
   },
+  env: {
+    ntaLayer: "dcp_nta_2010",
```
godfreyyeung, line 12: Maybe I should do NTA_LAYER

src/helpers/fetchNtaInfo.ts
```diff
@@ -0,0 +1,21 @@
+export const fetchNtaInfo = async (
+  ntacode: string | null,
+  onNtaInfo: (ntaInfo: any) => void
+): Promise<any> => {
+  if (!ntacode) return null;
+
+  const ntaInfoSql = `SELECT ntacode, ntaname FROM ${process.env.ntaLayer} WHERE ntacode='${ntacode}'`;
```
TylerMatteo, line 7: Do you need this separate query? I would think you already have access to this same data within the onClick handler in the Deck layer

godfreyyeung: Yeah, but it would require storing the data memory and passing it around components. Also was thinking about how the GeographyInfo component could be used outside the Map page, or loaded directly from URL and absent of a user click interaction.

TylerMatteo: Oh right, good call. I don't want to figure out how we would populate this data otherwise when the user comes directly to a URL

src/helpers/fetchNtaInfo.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  const {
+    rows: [ntaInfo],
+  } = await (
+    await fetch(`https://planninglabs.carto.com/api/v2/sql?q=${ntaInfoSql}`)
```
TylerMatteo, line 12: Couple little things:
Can you give this helper a return type so it's not `Promise<any>`?
What happens if that fetch call throws an error? I don't trust Carto's sql api to not randomly return 400 or 500 error codes :/

godfreyyeung: Good catch. I added some error handling for that.. It'll just return null and the UI will just display a blank geography

TylerMatteo (approved):

PR #71, opened 2022-03-02
6668 Add metatags for link preview

dhochbaum-dcp:
### Summary
Adds meta tags, as shown in Figma

Some of the screenshots still need to be added, and some of the current screenshots will need to be replaced.

#### Tasks/Bug Numbers
 - Completes [AB#6668](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/6668)

### DIFF
--- a/src/components/PageHeader.tsx
+++ b/src/components/PageHeader.tsx
@@ -43,35 +43,61 @@ export const PageHeader = () => {
       <meta
         name="title"
         content={`${currentPage.title} - NYC Equitable Development Data Tool`}
+        key="title"
       />
       <meta
         name="description"
         content="Explore New York City's housing conditions, demographic patterns, and more."
+        key="description"
       />
 
-      <meta property="og:type" content="website" />
-      <meta property="og:url" content="https://equitytool.planning.nyc.gov/" />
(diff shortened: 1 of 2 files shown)


godfreyyeung:
So the Next.Head is also already rendered in `src/pages/_document.tsx`. What about importing and rendering `<PageHeader />` in that page instead of `app.tsx`? Also the guides so suggest adding `key` attributes to each tag to future-proof against duplicate tags. 


image

dhochbaum-dcp:
> So the Next.Head is also already rendered in `src/pages/_document.tsx`. What about importing and rendering `<PageHeader />` in that page instead of `app.tsx`? Also the guides so suggest adding `key` attributes to each tag to future-proof against duplicate tags.
> 
> https://nextjs.org/docs/api-reference/next/head image

Putting <PageHeader> in document.tsx is not possible because hooks can only be called inside of the body of a function component, so you cannot call the useRouter hook.  Because of this, I placed it in app.tsx.  Unfortunately, we cannot remove the <Head /> from document.tsx, which requires it.

Thanks for pointing out the key - I've updated to add them.

godfreyyeung (approved):

PR #77, opened 2022-03-03
7069 Select sub-categories within a Data Table category

dhochbaum-dcp:
### Summary
When I am viewing a data table (for example, the Demographic Conditions, HEC, etc), I want to be able to toggle between various subcategories (racial and ethnic groups) to display the relevant data 

## Acceptance Criteria
UI For switching categories will be dropdown box for all screen sizes (this is different from previous AC where UI was different on desktop)
    Dropdown width should be full width (with padding) on sizes smaller than "md". When larger than md, size can be determined naturally by items inside the dropdown. Note that dropdown appears below the category menu on mobile and at the top of the main content section on desktop, this should be natural with how the page is already laid out. [Figma](https://www.figma.com/file/U7dgahwMaggEmMI1l5O1Z9/Equitable-Development-Reporting---EDDT-Tool?node-id=2701:48489)
    When user selects a subcategory, it should be reflected via a query parameter added to the url. Ex: /data/district/4004/demo?subgroup=bnp. We can use these codes for the groups:
        tot - Total Population
        wnh - White Non-hispanic
        bnh - Black Non-hispanic
        anh - Asian Non-hispanic
        hsp - Hispanic
    Default group when a user switches categories should be Total Population

#### Tasks/Bug Numbers
 - Finishes user story AB#6670 and task AB#7069

### DIFF
--- a/src/hooks/useDataExplorerState/useDataExplorerState.ts
+++ b/src/hooks/useDataExplorerState/useDataExplorerState.ts
@@ -1,22 +1,25 @@
 import { useRouter } from "next/router";
 import { Category } from "@constants/Category";
 import { Geography } from "@constants/geography";
+import { Subgroup } from "@constants/Subgroup";
 import { NYC } from "@constants/geoid";
 
 export interface DataExplorerState {
   geography: Geography;
   geoid: string;
   category: Category;
+  subgroup: string;
 }
 
 export const useDataExplorerState = (): DataExplorerState => {
   const router = useRouter();
-  const { geography, geoid, category } = router.query;
+  const { geography, geoid, category, subgroup } = router.query;
 
   const result: DataExplorerState = {
     geography: Geography.CITYWIDE,
     geoid: NYC,
     category: Category.DEMO,
+    subgroup: Subgroup.TOT,
   };
 
   if (
@@ -35,6 +38,13 @@ export const useDataExplorerState = (): DataExplorerState => {
     Object.values(Category).includes(category as Category)
   ) {
     result.category = category as Category;
+  }
+
+  if (
+    typeof subgroup === "string" &&
+    Object.values(Subgroup).includes(subgroup as Subgroup)
+  ) {
+    result.subgroup = subgroup as Subgroup;
   }
 
   return result;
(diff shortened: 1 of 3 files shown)


src/pages/data/[geography]/[geoid]/[category].tsx
```diff
@@ -222,6 +224,14 @@ const DataPage = ({ initialRouteParams }: DataPageProps) => {
   const [shouldShowReliability, setShouldShowReliability] =
     useState<boolean>(false);
 
+  const { geography, geoid, category, subgroup } = useDataExplorerState();
+  const router = useRouter();
+  const changeSubgroup = (event: any) => {
+    router.push(
+      `/data/${geography}/${geoid}/${category}?subgroup=${event.target.value}`
```
Comment (godfreyyeung) [inline src/pages/data/[geography]/[geoid]/[category].tsx:231]: Eventually we'll need to figure out how to use the `router.push()` API to tack on the subgroup, instead of asserting an absolute URL. There will be a showReliability queryParam eventually that shouldn't be overwritten

Comment (dhochbaum-dcp) [reply src/pages/data/[geography]/[geoid]/[category].tsx:231]: Wouldn't we add showReliability to useDataExplorerState and include it in the URL the same way I did with subgroup?

Comment (godfreyyeung) [reply src/pages/data/[geography]/[geoid]/[category].tsx:231]: Yeah that could work.. Though maybe we need to write some conditional logic to include or exclude query params from the URL depending on if null

src/hooks/useDataExplorerState/useDataExplorerState.ts
```diff
@@ -1,22 +1,25 @@
...
 import { Category } from "@constants/Category";
 import { Geography } from "@constants/geography";
+import { Subgroup } from "@constants/Subgroup";
 import { NYC } from "@constants/geoid";
 
 export interface DataExplorerState {
   geography: Geography;
   geoid: string;
   category: Category;
+  subgroup: string;
```
TylerMatteo, line 11: Are you able to make this be of type `Subgroup`?

dhochbaum-dcp: That appears to work, I've made the update.

TylerMatteo (approved):

PR #79, opened 2022-03-07
Refactor how layer visibility is handled to improve map performance

TylerMatteo:
### Summary
This is a quick PR to change how map layers are shown/hidden. Instead of changing the items of the `layers` array, this approach keeps the members of that array static and hides them with DeckGL's `visible` property, [as per their docs recommendation](https://deck.gl/docs/api-reference/core/layer#visible). I noticed a significant performance boost with this change, particularly when toggling between DRI and Data Tool "views". I also renamed the hook from `useSelectedLayer` to `useLayers`, to better describe what the hook actually does now.


godfreyyeung:
Curious if this affects first load performance?

TylerMatteo:
> Curious if this affects first load performance?

I just compared it visually with staging and it seems like it does slightly, for the reason you'd suspect - Deck pulls all 4 layers at once with this change. But I think the trade off is definitely worth it.

godfreyyeung:
> > Curious if this affects first load performance?
> 
> I just compared it visually with staging and it seems like it does slightly, for the reason you'd suspect - Deck pulls all 4 layers at once with this change. But I think the trade off is definitely worth it.

Yeah, it seems minimal. Though i'm on a beefy laptop, which makes it hard for me to tell what is a large load.

godfreyyeung (approved):

TylerMatteo:
> > > Curious if this affects first load performance?
> > 
> > 
> > I just compared it visually with staging and it seems like it does slightly, for the reason you'd suspect - Deck pulls all 4 layers at once with this change. But I think the trade off is definitely worth it.
> 
> Yeah, it seems minimal. Though i'm on a beefy laptop, which makes it hard for me to tell what is a large load.

Yeah definitely worth calling out at least - I guess I was subconsciously hoping Deck wouldn't load data for a layer until the first time it's visible is set to `true`. I'd imagine there's a way to do that if you need that much optimizations but apparently it isn't the default behavior. I think any performance hit here is going to be more bottlenecked to network speed rather than client machine speed.

PR #89, opened 2022-03-22
7482 Set up Data Explorer to pull and display Housing Production Data

TylerMatteo:
### Summary
This PR sets up app to pull and display data for Housing Production indicators

#### Tasks/Bug Numbers
 - Fixes [AB#7482](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/7482)

### Technical Explanation
This PR had to contain quite a few changes to meet requirements so I'll do my best to explain them here in an order that makes sense

#### Pulling and Validating data from Digital Ocean
I added a simple class called `DataExplorerService` to handle code related to downloading data from Digital Ocean via HTTP calls to the URLs where files are hosted. To use this class, just create an object that instantiates the class (`const myDataExplorerService = new DataExplorerService(baseURL)`. The constructor takes a single paramater - a string containing the base url where the files are located.

> ##### How does `DataExplorerService` know where to look?
> To see where the data files are located, go to the "Planning Labs" project in DO, then to a Space called `equity-tool-data`. In there you should see folders for each environment. The files themselves should be in those folders, with no additional nested folders. The files should follow the naming convention `<geography>-<geoid>-<category>.json`. "geography" should be one of `district`, `borough`, or `citywide`. Geoids for districts (pumas) should be the geoid for that puma (without the leading 0 that you will sometimes see), geoids for boroughs should be the 1 digit "borocode", and the only geoid for citywide should be `nyc`. Finally, <category> should be one of the 5 four character category codes used elsewhere in the app ("demo", "hopd", etc). **Note that the URL string passed to the constructor should include the piece with the environment folder and a trailing `/`, for example `https://equity-tool-data.nyc3.digitaloceanspaces.com/development/`**. Please download one of those files directly from DO yourself if you'd like an example of what one looks like structurally.
> 

DataExplorerService then creates an instance of an [Axios](https://axios-http.com/) client and has one method `get` that takes in the 3 pieces needed to get a file - geography, geoid, and category and returns a Promise representing the network call to download that file.

In order to validate that the returned data matches the schema we expect, I wrote some simple schema validators using the schema validating library [Yup](https://github.com/jquense/yup). We can use these schemas as a source-of-truth for the shape of JSON that our data processing should output. Making these schemas also has the added benefit of allowing us to easily create TS types for the data by using Yup's `InferType`. Those types are reused in component props types downstream. This schema may change as we add data for the other categories, but hopefully not too significantly.

#### Updating the Data Explorer page to use Static Site Generation and pull the data for given route params

First off, I made a small change to treat the "subgroup" (tot, anh, bnh, wnh, or hsp) in the route param as a segment param instead of an optional query param. This made more sense as I got into implementing the data fetching because this isn't really an "optional" param as far as the page is concerned. As a result, the file formally called `[category].tsx` is now located at `[subgroup].tsx`.

##### Changing `[subgroup].tsx` from `getServerSideProps` to `getStaticProps` and `getStaticPaths`
At a high level, Nextjs offers two ways of pre-rendering pages: Static Site Generate (SSG) and Server Side Generation (SSR). Pages that do SSR will _always_ export a function called `getServerSideProps`. Pages that do Static Site Generation will _always_ export a function `getStaticProps` and _optionally_ export a function called `getStaticPaths`. [This article](https://vercel.com/blog/nextjs-server-side-rendering-vs-static-generation) by the authors of Nextjs explains the differences between SSG and SSR conceptually as well as how they're implemented in Nextjs better than I can so have a look if you'd like an overview.

The `[subgroup].tsx` page had exported a `getServerSideProps` function as some boilerplate to have in place until we decided if SSG or SSR made more sense. Given that this page only ever shows data for one combination of geography, geoid, category, and subgroup, and given that the total unique combinations of geography, geoid, category, and subgroup that we need have data for is relatively low, I decided to switch the page over to SSG to take advantage of the improved performance and UX that it offers. The function `getStaticPaths` handles enumerating all of those combinations (using pumas.json to get the list of pumas). The object it returns has a property `paths`, which is an array of the valid paths for this page (ex `["/borough/1/hopd/tot","/district/3701/demo/anh", etc...]`). This means that the site will show a 404 page if given a URL with invalid values for any of the route segment params. `getStaticProps` contains code for downloading the data for a given site of values for the route params. Because the data files on DO contain all data for given geoid and category, this function will pull down that data, and return only the part of the data file corresponding to the subgroup in the route params. The `props` returned in `getStaticProps` are the props given to the component for the page (`DataPage` in this case).

#### New Components
I added some new components that iterate over the returned data to display tables. While I took some code from them, I deleted `DataTable` and `EstimateTable` in favor of the new equivalent `VintageTable`. My intention is that `VintageList` and `VintageTable` will handle all of the responsive table requirements but for some of that work will come in future PRs because the Housing Production tables only ever have one vintage per indicator and no CoT tables. The hierachy of how then new components are used can be described as: `[subgroup].tsx` renders a list of `<Indicator>`s, which renders a list of `<VintageList>`s, which renders a list of `<VintageTable>`s, which finally renders several lists of `<DataPointCells>`. The prop types for these component make use of the Type generated by the Yup schemas

Screenshots taken at `data/borough/1/hopd/tot`:
image

image


cypress/integration/pages/data.geography.geoid/category.spec.ts
```diff
@@ -1,6 +1,6 @@
 describe("data/geography/geoid/category page", () => {
   beforeEach(() => {
-    cy.visit("data/district/4001/demo");
+    cy.visit("data/borough/1/hopd/tot");
```
godfreyyeung, line 3: Thanks for this!

src/hooks/useLayers/useLayers.ts
```diff
@@ -84,20 +84,20 @@ export const useLayers = (): CartoLayer<any, any>[] | null => {
       data: `SELECT * FROM dcp_borough_boundary`,
       uniqueIdProperty: "id",
       getLineColor: (feature: any) => {
-        if (feature?.properties?.boroname?.trim() === geoid?.trim()) {
+        if (feature?.properties?.borocode?.toString() === geoid?.trim()) {
```
godfreyyeung, line 87: Thanks for this

godfreyyeung:
Thanks for the thorough wonderful writeup! :) 🪄  
One suggestion is to use boroname instead of borocode, I think that'll make things more convenient and make it easier to communicate about the files in the future.

Love capturing the schema of each data slice as frontend objects. 

Yes, prefer smaller PRs -- like maybe the tables could come in another PR. 

Hope we can get those tests running again soon. There's probably some extra setup for SSG during e2e tests.

godfreyyeung (approved):

PR #91, opened 2022-03-23
7511 Implement Google Analytics event tags

dhochbaum-dcp:
### Summary
This adds most of the event tags necessary for tracking in Google Analytics.

#### Tasks/Bug Numbers
 - Partially completes AB#7511

### DIFF
--- a/src/components/CategoryMenu/CategoryMenu.tsx
+++ b/src/components/CategoryMenu/CategoryMenu.tsx
@@ -9,6 +9,7 @@ import {
   faUmbrella,
   faHouseUser,
 } from "@fortawesome/free-solid-svg-icons";
+import ReactGA from "react-ga4";
 
 export interface CategoryMenuProps extends FlexProps {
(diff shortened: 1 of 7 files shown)


src/components/CategoryMenu/CategoryMenu.tsx
```diff
@@ -33,34 +34,69 @@ export const CategoryMenu = ({
       icon={<FontAwesomeIcon icon={faUserGroup} />}
       href={`/data/${geography}/${geoid}/${Category.DEMO}`}
       isActive={currentCategory === Category.DEMO}
+      onClick={() => {
+        ReactGA.event({
```
TylerMatteo, line 38: Just curious - are you able to test this onClick in a preview deployment and make sure the event shows up in GA? I only ask for this scenario because these links send users to a different page, which makes me wonder if the onClick could be getting cut off before it has a chance to complete.

TylerMatteo (approved):

PR #96, opened 2022-03-25
7627 Add category text/icon displaying risk bins

dhochbaum-dcp:
### Summary
for each NTA, in addition to an overall displacement risk value (highest, higher... <---> ...lower, lowest), each of the subindices/categories (Vulnerability, Housing, Market Pressure) also have a similar value associated. These values will be displayed next to their respective titles in the sidebar when a user selects a given NTA 

#### Tasks/Bug Numbers
 - Completes [AB#7627](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/7627)


src/components/SidebarContent/SidebarContent.tsx
```diff
@@ -10,11 +10,203 @@ import { useClearSelection } from "@helpers/useClearSelection";
...
 import { Geography } from "@constants/geography";
 import { CategoryMenu } from "@components/CategoryMenu";
+import { SubindicatorBin } from "@components/SidebarContent";
 
 export const SidebarContent = () => {
   const { view, geoid, geography } = useMapSubrouteInfo();
   const clearSelection = useClearSelection();
 
+  const ntaIndexes: { [index: string]: any } = {
+    BK29: "Highest",
```
TylerMatteo, line 20: Where do all of these come from? I don't think we should have data inlined in components like this.

dhochbaum-dcp: Moved to a json.

TylerMatteo (approved):

PR #98, opened 2022-03-29
6875 Add functionatlity to Vintage components for handling multiple vintages

TylerMatteo:
### Summary
This PR does three main things:

* Adds some functionality to the Indicator and Vintage components for handling the styling on desktop and mobile when there are multiple vintages under an indicator. 
* Updates the `getStaticPaths` in [subgroup].tsx to only attempt to pull Housing Security, Affordability, and Quality data - there is data in DO for 3 indicators in that category as of the writing of this PR. It should be the first category fully processed
* Updates the `dataPoint` schema to match some recent changes to our data pipeline and updates `DataPointCell.tsx` accordingly

One thing this PR _doesn't do_ that may seem to be missing: hooking up `shouldShowReliability` and adding logic to Vintage components for hiding those cells when it is set to true. In this PR, indicators with cv, moe, etc, will always show those cells. I'm going to cover that in another upcoming PR to keep the scope of this PR (somewhat) reasonable.

#### Tasks/Bug Numbers
 - Fixes [AB#6875](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/6875)

#### Rendering multiple "Vintages"
When we have data for an indicator for more than one point in time, we render multiple "Vintages". On Mobile, those are stacked vertically:
image

And desktop those are "stacked" horizontally:
image

`VintageTable.tsx` is responsible for rendering the data cells for one Vintage (each screenshot above is showing 2 `VintageTables`). `VintageList` is responsible for rendering each Vintage in the indicator and handles [stacking them in the correct orientation based on screen width](https://github.com/NYCPlanning/equity-tool/compare/task/6875-explorer-desktop-styling?expand=1#diff-bc7d98451e76a80f04ef17ada7997884e73537ec741abc42c9cd1f1701145089R33). However, when they're stacked horizontally, we only need to show the "row labels" ("Total population", "Asian non-Hispanic", etc) once - for the first Vintage, so `VintageTable` has [some styling](https://github.com/NYCPlanning/equity-tool/compare/task/6875-explorer-desktop-styling?expand=1#diff-00f99adfa4a0384bb6065db6fd921960de2cc6205411eaa34e08289caf7eef88R45) to hide those cells on Vintages after the first. However, doing that causes the row heights to change when the row labels are long enough to take up more than one line, so rows don't line up properly:
image

To address this, I added [the ref and `rowHeights` in VintageList](https://github.com/NYCPlanning/equity-tool/compare/task/6875-explorer-desktop-styling?expand=1#diff-bc7d98451e76a80f04ef17ada7997884e73537ec741abc42c9cd1f1701145089R22) which passes a ref to the first vintage, stores each row's height, and sets those heights on the rows of the following vintages. I realize this solution feels a little heavy handed but it fixes the only thing stopping me from having to build completely separate desktop and mobile table components and I think this solution is preferable to that.

#### Updates to `dataPointSchema`
I updates `dataPointSchema` to have properties `measure` and `vintage`, with the list of possible values for each - this gives us a better way to encode "metadata" about the datapoint and matches with Data Engineering's terminology.

### DIFF
--- a/src/components/DataPointRow/DataPointRow.tsx
+++ b/src/components/DataPointRow/DataPointRow.tsx
@@ -1,0 +1,44 @@
+import { Tr, Td, TableRowProps } from "@chakra-ui/react";
+import { Row } from "@schemas/row";
+import { DataPointCell } from "@components/DataPointCell";
+
+export interface DataPointRowProps extends TableRowProps {
+  row: Row;
+}
+
+export const DataPointRow = ({ row, ...props }: DataPointRowProps) => {
+  const { label, isDenominator, cells } = row;
+  return (
+    <Tr
+      {...props}
+      _last={{
+        th: {
+          borderBottomLeftRadius: { base: "0.75rem", md: "0rem" },
+        },
+        "td:last-of-type": {
+          borderBottomRightRadius: {
+            base: "0.75rem",
+            md: "0rem",
+          },
+        },
+      }}
+    >
+      <Td
+        as="th"
+        scope="row"
+        fontWeight={isDenominator ? "700" : "400"}
+        zIndex={"100"}
+        minWidth={{ base: "calc((100vw - 26px) / 3)", md: "unset" }}
+        maxWidth={{ base: "calc((100vw - 26px) / 3)", md: "unset" }}
+        px={{ base: "0.375rem", md: "1.5rem" }}
+        position={"sticky"}
+        left={"0"}
+      >
+        {label}
+      </Td>
+      {cells.map((dataPoint, j) => (
+        <DataPointCell key={`data-point-cell-${j}`} dataPoint={dataPoint} />
+      ))}
+    </Tr>
+  );
+};
--- a/src/pages/data/[geography]/[geoid]/[category]/[subgroup].tsx
+++ b/src/pages/data/[geography]/[geoid]/[category]/[subgroup].tsx
@@ -98,7 +98,11 @@ export const getStaticProps: GetStaticProps = async (context) => {
     const res = await dataExplorerService.get(geography, geoid, category);
     // Validate file follows expected schema
     const profile = await categoryProfileSchema.validate(res.data);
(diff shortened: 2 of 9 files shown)


src/pages/data/[geography]/[geoid]/[category]/[subgroup].tsx
```diff
@@ -241,13 +248,13 @@ const DataExplorerNav = () => {
 
 const DataPage = ({ hasRacialBreakdown, indicators }: DataPageProps) => {
   const [shouldShowReliability, setShouldShowReliability] =
-    useState<boolean>(false);
-
+    useState<boolean>(true);
+  console.log(hasRacialBreakdown);
```
Comment (godfreyyeung) [inline src/pages/data/[geography]/[geoid]/[category]/[subgroup].tsx:252]: is this necessary?

src/pages/data/[geography]/[geoid]/[category]/[subgroup].tsx
```diff
@@ -78,6 +78,7 @@ export const getStaticPaths: GetStaticPaths = () => {
 };
 
 export const getStaticProps: GetStaticProps = async (context) => {
+  console.log("foo");
```
Comment (godfreyyeung) [inline src/pages/data/[geography]/[geoid]/[category]/[subgroup].tsx:81]: ahhh

src/components/DataPointRow/DataPointRow.tsx
```diff
@@ -0,0 +1,44 @@
...
+        as="th"
+        scope="row"
+        fontWeight={isDenominator ? "700" : "400"}
+        zIndex={"100"}
+        minWidth={{ base: "calc((100vw - 26px) / 3)", md: "unset" }}
+        maxWidth={{ base: "calc((100vw - 26px) / 3)", md: "unset" }}
+        px={{ base: "0.375rem", md: "1.5rem" }}
+        position={"sticky"}
+        left={"0"}
+      >
```
godfreyyeung, line 36: these should just be strings, without the curlies, right?

godfreyyeung (approved):

PR #99, opened 2022-03-29
Make e2e tests compatible with SSG

godfreyyeung:
I initially tried using `cy.intercept()`, but I think the problem with that is Static Site Generation runs `getStaticProps()` _before_ tests, so the requests to Digital Ocean for static files won't be intercepted. The workaround here is to set up a fixture folder within `src` and have in-app logic to toggle use of the fixtures, instead of inside the cypress framework.

One advantage of having the  `USE_FIXTURE` toggle in-app is that devs could potentially leverage it to continue loading the Data Page for UI development during data churn. Devs just need to watch out for schema changes.


godfreyyeung:
Closing for now. We should get the data in `/development` stable-ish

PR #102, opened 2022-03-30
7549 add CSS to control button size and background per figma designs

horatiorosa:
### Summary

   - The site navigation hamburger menu (HM) on mobile screens, specifically on Safari browsers was taking 50% of the header real estate. The HM was also rendering with grey and dark grey (when the menu was selected and active) contrary to designs.
   - I added CSS to to set a max-width on the HM and removed the background colors
   #### before fix
hg-mobile-inactive
hg-mobile-active

#### after fix
hg-mobile-fixed

#### Tasks/Bug Numbers
 - Fixes [AB#7549](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/7549)

[AB7549](https://dev.azure.com/NYCPlanning/ITD/_workitems/edit/7549)

### DIFF
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -62,11 +62,11 @@ export const Header = () => {
               w={"100%"}
               h={"100%"}
               color={isOpen ? "teal.600" : "gray.600"}
+              maxWidth={6}
             />
           }
           display={{ base: "flex", md: "none" }}
           onClick={onToggle}
-          backgroundColor={"white"}
           minWidth={6}
           h={4}
           _hover={{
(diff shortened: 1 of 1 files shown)


src/components/Header/Header.tsx
```diff
@@ -80,6 +80,12 @@ export const Header = () => {
           }}
           mr={6}
           transform={`scale(${isOpen ? "1.125, 1.0833" : "1, 1"})`}
+          bg="none"
+          _active={{
+            bg: "none",
+            minWidth: 7,
```
godfreyyeung, line 86: Curious what this minWidth accomplishes?

godfreyyeung:
I noticed that on click, during that brief "active" state, the hamburger icon shrinks to a tiny icon. Tested in Chrome 
image
Not extremely critical at this point but curious whether that happened before

src/components/Header/Header.tsx
horatiorosa: This was in the ADO ticket (a new minWidth for the button when it's active) but it was causing the bounce effect (small, then large again) on the hamburger so I had to remove it.

godfreyyeung (approved):

PR #104, opened 2022-03-31
7777 changed heading  in the Community Data sidebar to match figma reqs

horatiorosa:
Heading in the Community Data sidebar was incorrect, updated h2 to h3 and set font size to 25px per figma

fixes AB#7777


src/components/WelcomeContent.tsx
```diff
@@ -1,14 +1,15 @@
 import { Heading, Text, Link } from "@chakra-ui/react";
 import { useMapSubrouteInfo } from "@hooks/useMapSubrouteInfo";
-import ReactGA from "react-ga4";
```
godfreyyeung, line 3: Was this an intentional delete?

src/components/WelcomeContent.tsx
```diff
@@ -1,14 +1,15 @@
...
 
 const WelcomeContent = () => {
   const { view } = useMapSubrouteInfo();
 
   if (view === "datatool") {
     return (
       <>
-        <Heading>Welcome!</Heading>
+        <Heading as="h3" fontSize="25px">
```
godfreyyeung, line 10: We are trying to use rem instead of px throughout the app. Here you can divide 25 by 16 to get the rem value `1.5625rem`
The figma specified 25px as font size? interesting.

horatiorosa: ohh, ok, i will make a note of that. I wasn't sure what was out base rem unit, since it's 16, i'll use that in my current PR

godfreyyeung: Ok!

godfreyyeung (approved):

PR #135, opened 2022-04-08
7917 fixed overlapping text bug on smaller height browser windows

horatiorosa:
- moved welcome footer text to welcome content
 - enabled Y scrolling
 - cleaned up unused code in welcome context after refactoring

#### Tasks/Bug Numbers
 - Fixes [AB#7917](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/7917)

### Technical Explanation
The height for the `WelcomeContext` component was set to 100%  forcing the component to remain the same height at all vertical browser heights  which results in the `WelcomeFooter` text to overlap into and over the `WelcomeContext` text. By importing the `WelcomeFooter` into the `WelcomeContext` component rather than have them individually imported into the `SidebarContent` component, the overlapping text bug is avoided and the entire body of text scrolls as one unit. I also added a scrollbar to visually clue in the user that there is scrollable text.

### DIFF
--- a/src/components/SidebarContent/SidebarContent.tsx
+++ b/src/components/SidebarContent/SidebarContent.tsx
@@ -3,7 +3,6 @@ import { Box, Divider, Button, Flex } from "@chakra-ui/react";
 import { ArrowBackIcon } from "@chakra-ui/icons";
 import { GeographyInfo } from "@components/GeographyInfo";
 import WelcomeContent from "@components/WelcomeContent";
-import WelcomeFooter from "@components/WelcomeFooter";
 import { DRMSelection } from "@components/SidebarContent/DRMSelection";
 import { useMapSubrouteInfo } from "@hooks/useMapSubrouteInfo";
 import { useClearSelection } from "@helpers/useClearSelection";
@@ -76,14 +75,12 @@ export const SidebarContent = () => {
     <>
       <Box
         height="100%"
+        overflowY="auto"
         justify="space-between"
         padding="2rem 1rem"
         paddingBottom="0rem"
       >
         <WelcomeContent />
-      </Box>
-      <Box padding="2rem 1rem" paddingTop="0rem">
-        <WelcomeFooter />
       </Box>
     </>
   );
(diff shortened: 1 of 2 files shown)


godfreyyeung:
Sorry I should have flagged this earlier, but the original intent of having the content and footer in separate components is so that the footer can be aligned to the bottom of the side panel, per original specs. Trying to think of simple way to have it be aligned at the bottom when there's enough vertical space, but sit in-flow with the content when the screen height shrinks... 
But if design is okay with this change (and the footer no longer aligning at the bottom), then i'm ok with this change

godfreyyeung:
Alright, nevermind. I'll table that for another day, since design weighed in and agreed to this scrolling.

src/components/SidebarContent/SidebarContent.tsx
```diff
@@ -76,15 +75,13 @@ export const SidebarContent = () => {
     <>
       <Box
         height="100%"
+        overflowY="auto"
         justify="space-between"
```
godfreyyeung, line 79: I think you can get rid of this `justify`, it actually wasn't useful and isn't used anymore

godfreyyeung (changes requested):
Unfortunately here it won't work to add WelcomeFooter into WelcomeContent. They're meant to be sibling components. Currently both are rendered in MobileDrawer, so there's a duplicate footer.
image

Here you have to either remove <WelcomeFooter /> from MobileDrawer to prevent a duplicate footer, or instead render WelcomeFooter sibling to WelcomeContent in SidebarContent. 

```
      <Box
        height="100%"
        overflowY="auto"
        justify="space-between"
        padding="2rem 1rem"
        paddingBottom="0rem"
      >
        <WelcomeContent />
        <br />
        <br />
        <br />
        <WelcomeFooter />
      </Box>
```
Let's hold off on merging and pair a bit on this first

src/components/SidebarContent/SidebarContent.tsx
horatiorosa: 👍

horatiorosa:
@godfreyyeung  dammit, you're right. I didn't realize there was a mobile drawer, my bad. Sorry. It seems to be a quick fix by tweaking the display attribute:
```
<Box paddingTop="2rem" display={{ base: "none", md: "block" }}>
     <WelcomeFooter />
</Box>
```

What do you think?

PR #143, opened 2022-04-12
7920 Expand and Collapse all tables on mobile

godfreyyeung:
Fixes [AB#7920](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/7920)

Introduces Expand All and Collapse All (table) buttons to the mobile view of the Data Explorer page. 

Here I made use of Contexts to avoid passing functions through multiple levels from `[subgroup]` down to `VintageTable`. 

The Context essentially acts as a binding between the state array `tableSetIsOpens` in `[subgroup]` and the `setIsOpen` Dispatches of each table.  `tableSetIsOpens` holds an array of `setIsOpen` dispatches from the tables. 

When a table is rendered, it will use the Context's `addSetIsOpen: ()` property  to push it's `setIsOpen` dispatch to the `tableSetIsOpens` array. 

image

### DIFF
--- a/src/components/VintageTable/VintageTable.tsx
+++ b/src/components/VintageTable/VintageTable.tsx
@@ -1,19 +1,10 @@
-import { forwardRef } from "react";
-import {
-  Flex,
-  Box,
-  Text,
-  Table,
-  Thead,
-  Tbody,
-  Th,
-  Tr,
-  useDisclosure,
-} from "@chakra-ui/react";
+/* eslint-disable react/display-name */
+import { forwardRef, useState } from "react";
+import { Flex, Box, Text, Table, Thead, Tbody, Th, Tr } from "@chakra-ui/react";
 import { ChevronDownIcon } from "@chakra-ui/icons";
 import { DataPointRow } from "@components/DataPointRow";
 import { Vintage } from "@schemas/vintage";
-
+import TablesIsOpenContext from "@contexts/TablesIsOpenContext";
 export interface VintageTableProps {
   vintage: Vintage;
   rowHeights: number[];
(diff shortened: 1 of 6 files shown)


src/components/VintageTable/VintageTable.tsx
```diff
@@ -24,176 +15,217 @@ export interface VintageTableProps {
...
+                      md: "table-row",
+                    }}
+                  >
+                    <Th
+                      rowSpan={headers.length}
+                      display={{ base: "table-cell", md: "none" }}
+                      position={"sticky"}
+                      left={"0"}
+                      zIndex={"100"}
+                      minWidth={{
```
godfreyyeung, line 131: My prettier updated these, probably due to the entire block being indented.

src/components/VintageTable/VintageTable.tsx
```diff
@@ -24,176 +15,217 @@ export interface VintageTableProps {
...
-              }}
-              px={"1rem"}
-              minWidth={{
-                base: "calc(100vw - 26px)",
-                md: isSurvey && shouldShowReliability ? "30.5rem" : "15.375rem",
-              }}
-              maxWidth={{
-                base: "calc(100vw - 26px)",
-                md: isSurvey && shouldShowReliability ? "30.5rem" : "15.375rem",
+      <TablesIsOpenContext.Consumer>
```
TylerMatteo, line 21: I think it's best to avoid rendering `Consumer`s. The preferred way of accessing a context is via the `useContext` hook. So instead of having these nested return statement in your TSX, you can keep the return statement in this component the same as it was before and do
```
const { addSetIsOpen } = useContext(TablesIsOpenContext);
addSetIsOpen(setIsOpen);
```
just before the return (on line 19 or so).

src/components/VintageTable/VintageTable.tsx
```diff
@@ -1,19 +1,10 @@
...
-  Box,
-  Text,
-  Table,
-  Thead,
-  Tbody,
-  Th,
-  Tr,
-  useDisclosure,
-} from "@chakra-ui/react";
+/* eslint-disable react/display-name */
```
TylerMatteo, line 1: I would have thought changing the consumer to a `useContext` change I described below would fix this but I tried it and it looks like that's not the case. That leads me to believe it might be related to my usage of `forwardRef`, though Idk why it's only showing up now. In any case, having `displayName` be set to something is important for debugging purposes, and so is avoiding turning off our linting rules. I think the best solution here is simply to set the `displayName` explicitly. To do that, you can change line 15 `export const VintageTable = forwardRef<HTMLTableElement, VintageTableProps>` to `const VintageTable = forwardRef<HTMLTableElement, VintageTableProps>`, then add the display name after the component definition with `VintageTable.displayName = "VintageTable";` and finally, export the component `export { VintageTable };`

src/contexts/TablesIsOpenContext.ts
```diff
@@ -0,0 +1,9 @@
+import { createContext } from "react";
+
+const TablesIsOpenContext = createContext({
+  addSetIsOpen: (dispatch: React.Dispatch<boolean>): void => {
+    console.log(dispatch);
```
TylerMatteo, line 5: The change needed to get rid of this "placeholder" console.log actually gets at a very commonly confusing thing about contexts - why do I need to pass a "default" value to `createContext` and a `value` prop to the Provider? Basically, the value you pass in to `createContext` is what will be returned to Consumers _when you try to access the context without being inside of a Provider_. Usually, if that happens it means something has gone wrong. The best way I've come across to solve this while also making your context provider code more expressive, is detailed [here](https://kentcdodds.com/blog/how-to-use-react-context-effectively) but essentially you add an `| undefined` to the type param you pass to `useContext` and pass in undefined here, instead of a "placeholder" value, and add a check for undefined when you use the context.

It's a relatively straightforward change to make but I'll defer to you if you think it's necessary given the work it would take. If you'd rather hold off on that refactoring, I'd just ask that we avoid leaving console.logs in the code in these scenarios. You could instead change line 5 to just `return;` which will result in an eslint warning about no-unused-vars which can be silenced by adding `/* eslint-disable @typescript-eslint/no-unused-vars */` to the top of the file. Obviously that's no ideal either but I think it's better than potentially printing stuff to users' browser consoles.

src/contexts/TablesIsOpenContext.ts
```diff
@@ -0,0 +1,9 @@
...
+  },
+});
+
+export default TablesIsOpenContext;
```
TylerMatteo, line 9: Not sure if we ever 100% agreed on a best practice for this but I'm still in favor of doing named exports over defaults when possible

src/components/VintageTable/VintageTable.tsx
godfreyyeung: Gotcha, yeah I was ambivalent about which style to use. This Consumer component style is "cuter"  but `useContext` is more readable for sure. Since you prefer that too, let's definitely go with that style

src/components/VintageTable/VintageTable.tsx
godfreyyeung: ahh. I was trying to add displayName to my context. Thanks! will try this.

src/contexts/TablesIsOpenContext.ts
godfreyyeung: i'll change it back, I was seeing if it helped with the displaname type error

src/contexts/TablesIsOpenContext.ts
godfreyyeung: I'll try out the `undefined` method. 

 I don't like console.logs either. So far it seems that's how [we've been avoiding "unused variable" type errors in the app](https://github.com/NYCPlanning/equity-tool/blob/main/src/pages/map/%5Bview%5D/%5Bgeography%5D.tsx#L87), so I defaulted to this for consistency. I'd rather stay consistent than to use `return` and silence it.

src/contexts/TablesIsOpenContext.ts
godfreyyeung: I think we agreed to use named exports over defaults

src/components/VintageTable/VintageTable.tsx
TylerMatteo: Sounds good! Yeah it's on of those "old React" vs "new React" things that no one ever tells you explicitly when you're learning it. My understanding is that the Consumer with "render prop" way of doing it was how you did in the "pre-hooks" world.

src/contexts/TablesIsOpenContext.tsx
```diff
@@ -0,0 +1,41 @@
...
+          tablesSetIsOpens.push(setIsOpen);
+        },
+      }}
+    >
+      {children}
+    </TablesIsOpenContext.Provider>
+  );
+}
+
+function useTablesIsOpen() {
```
TylerMatteo, line 29: +1 for making a custom "wrapper" hook!

TylerMatteo (approved):

TylerMatteo (approved):

PR #149, opened 2022-04-12
8040 Accomodate SSG during tests with longer step time allowance

godfreyyeung:
Fixes [AB#8040](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/8040)

This fixes the tests by increasing the timeout for `start-server-and-test`, the utility used by Cypress to wait for a connection to localhost:3000.  Tests have been failing in Github Actions because SSG ran over the default timeout of start-server-and-test. 

See https://github.com/bahmutov/start-server-and-test

It also fixes some regressions in the geography spec due to copy and formatting changes.

### DIFF
--- a/.github/workflows/feature_branch.yml
+++ b/.github/workflows/feature_branch.yml
@@ -92,6 +92,7 @@ jobs:
       NEXT_PUBLIC_MAPBOX_TOKEN: ${{ secrets.MAPBOX_TOKEN }}
       NEXT_PUBLIC_DO_SPACE_URL: ${{ secrets.DO_SPACE_URL }}
       SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+      WAIT_ON_TIMEOUT: 900000
     steps:
       - uses: actions/checkout@v2
       - name: Setup node env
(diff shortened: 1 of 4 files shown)


horatiorosa (approved):
Seems like a decent amount time to wait to run the e2e

godfreyyeung:
> Seems like a decent amount time to wait to run the e2e

Yeah maybe. Tyler said one category took 7 minutes in itself. If we have 5 categories, that's upwards of 35 minutes. We can definitely trim this down if it ends up being shorter than 30 minutes on average

godfreyyeung:
> Seems like a decent amount time to wait to run the e2e

According to the tests, they ran in ~7 minutes. So I adjusted the timeout to be around double that for good measure: around 15 minutes.

TylerMatteo:
>

TylerMatteo:
> Seems like a decent amount time to wait to run the e2e

Yeah that definitely isn't normal but I'm fine with merging in this solution for now. I'm pretty sure the way you're supposed to handle this situation is to not do SSG for your e2e tests and have the pages built at run-time, like they are when you're doing local development. I'd like for us to get back to how the action workflows were originally implemented where they wouldn't continue to deployments for staging and prod if e2e tests failed but I don't think it's practical to do that until we figure out how to not need this timeout.

TylerMatteo (approved):
We should revisit how to run these tests without having to SSG the whole site but I think this is a great improvement over the current state of things.

godfreyyeung:
> We should revisit how to run these tests without having to SSG the whole site but I think this is a great improvement over the current state of things.

Sounds great. Thanks!

PR #155, opened 2022-04-15
8003 Prevent redundant view toggling

godfreyyeung:
- Fixes AB#8003

This PR prevents a user from re-clicking  the the toggle for the current view. It also adds guards to the toggle event handlers to make sure that any route processing -- and setting of previous selected geography -- only occurs if user is switching to the other view.

### DIFF
--- a/src/pages/map/[view]/[geography].tsx
+++ b/src/pages/map/[view]/[geography].tsx
@@ -111,48 +111,52 @@ const MapPage = ({ initialRouteParams }: MapPageProps) => {
   const [lastBoroughGeoid, setLastBoroughGeoid] = useState<string | null>(null);
 
   const onDrmClick = () => {
-    setLastCommunityDataGeography(geography);
-    setLastCommunityDataGeoid(geoid);
+    if (view === "data") {
+      setLastCommunityDataGeography(geography);
+      setLastCommunityDataGeoid(geoid);
 
-    ReactGA.event({
-      category: "Toggle Tool",
-      action: "Click",
-      label: "Displacement Risk Map",
-    });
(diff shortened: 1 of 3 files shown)


godfreyyeung:
@dhochbaum-dcp  Do you think it's appropriate to move the Google Analytics tracking inside the `if ( view === ...)` statements?

src/pages/map/[view]/[geography].tsx
```diff
@@ -111,48 +111,52 @@ const MapPage = ({ initialRouteParams }: MapPageProps) => {
   const [lastBoroughGeoid, setLastBoroughGeoid] = useState<string | null>(null);
 
   const onDrmClick = () => {
```
Comment (TylerMatteo) [inline src/pages/map/[view]/[geography].tsx:113]: Instead of wrapping the contents of this click handlers in if blocks, do you think it would be feasible to refactor `onDrmClick` and `onCommunityDataClick` into one function with a name like `onViewToggle` that would take "drm" or "data" as an argument? That way you could conditionally render the button for the currently active view to not have an onClick at all or to have one that is a no-op.

Comment (godfreyyeung) [reply src/pages/map/[view]/[geography].tsx:113]: I could. Are there any functional benefits to doing that? Or would it be mostly a style preference?

Comment (TylerMatteo) [reply src/pages/map/[view]/[geography].tsx:113]: Just stylistic. Assuming everything is inside the existing if blocks (including GA events), I don't think it would have any significant functional effects.

godfreyyeung:
@horatiorosa wrapping this up. Closing in favor of https://github.com/NYCPlanning/equity-tool/pull/162#pullrequestreview-1015532291

PR #156, opened 2022-04-15
8012 Hide back button on citywide geography

godfreyyeung:
Fixes [AB#8012](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/8012)
When a user has selected the citywide geography tab, there should not be a "back" button because the citywide geography is always selected by default under the citywide tab. Here, we keep the back button in the sidebar flow to prevent sidebar contents from jumping up and down between geography tabs. We only hide it by assigning `visibility: hidden` when citywide geography is selected.


horatiorosa (approved):
LGTM.
🚀 🚀

PR #157, opened 2022-04-16
experimenting with pom and custom commands

ileoyu:
Experimenting with Page Object Model (POM) design to reduce code redundancy and improve test case maintenance. 

Although POM adds a layer of complexity and isn't recommended by the Cypress team as best practices, I think the benefits still outweigh the cons. A major benefit that POM provides is that test cases are easier to fix, maintain, and insured from the brittleness due to the nature of its fast changing UI components, DOM elements, contents, etc. 

Added new custom Cypress commands to navigate on/through the map and to make assertions on downloaded files. 

Modified Cypress's default implicit wait time (from 4 sec to 15) so that assertions aren't prone to failure from slow loading DOM and allows time for fetch requests to complete.


cypress/integration/pages/pageObjectModel/communityDataMapDrawer.js
```diff
@@ -0,0 +1,32 @@
+class communityDataMapDrawer {
+
+    get geoInfoName() {
```
TylerMatteo, line 3: This isn't a huge deal but I don't think using getters here is really best practice - getters are usually for when you want to encapsulate access to an otherwise private member of the class. These functions don't access members (properties) of the class. I think simple methods like `getGeoInfoName` would suffice.

cypress/integration/pages/pageObjectModel/download.js
```diff
@@ -0,0 +1,29 @@
+class download { 
+
+    get downloadDataBtn() {
+        return cy.get('.css-zlmdk2 > .chakra-button')
```
TylerMatteo, line 4: In general, don't use class names for selection. Using them for grabbing specific bits of the DOM is generally considered a bad practice because it makes them responsible for more than just assigning styles to the element. Plus, these random string classes are liable to change unpredictably. Instead, we can use [set up RTL's Cypress plugin and use that](https://testing-library.com/docs/cypress-testing-library/intro/) or just add some `data-cy` attributes if you need something to select on.

cypress/integration/pages/pageObjectModel/dataExplorerHeader.js
```diff
@@ -0,0 +1,24 @@
...
+class subGroupDropDown { 
+
+    get anh() {
+        return cy.get('.chakra-select').select('anh')
+    }
+
+
+}
+
+export default new subGroupDropDown
```
TylerMatteo, line 22: Did you find this `export default new` pattern in a guide or resource somewhere? I've just never seen it before and it strikes me as a little redundant. Typically, files that declare classes would export the class itself, whereas I'm pretty sure this is going to export an instance of the class. If you just need one instance, why not just export a regular object?

cypress/integration/pages/Map/view/DRI_Subindices_Test.js
```diff
@@ -0,0 +1,19 @@
+import data_in from "./DRI_Subindices_Indicators_test.json";
```
TylerMatteo, line 1: Do you still need this file? It doesn't seem to test anything.

cypress/integration/pages/data.geography.geoid.category/pom&commandExperiment.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+      download.xls; // selects .xls file type 
+
+      download.downloadDataModalBtn.click(); // clicks on 'Download Data' from Modal 
+
+      cy.verifyDownload('4001_xlsx.zip', {timeout:5000}); // assert that the correct file is downloaded
+      
+    });
+
+    it.only('should click on drawer content', () => {
+      //cy.tableTesting('district', '4001', 'hsaq', 'tot');
```
TylerMatteo, line 76: Remove this?

TylerMatteo (changes requested):
Happy to talk through any of my comments. In general, I don't have problem with using the POM pattern if that's what you prefer. It's a pretty well established pattern for testing and seems to get the job done when implemented properly. I would just suggest that maybe at some point (even on a future project), you try writing some tests with the [App Actions](https://www.cypress.io/blog/2019/01/03/stop-using-page-objects-and-start-using-app-actions/) pattern that cypress recommends and share the pros and cons you see with each.

One other thing to keep in mind for future work is that I'd like to see our integrations tests written in Typescript. It looks like Cypress has pretty good support for it and it'll just help with keeping things consistent.

PR #159, opened 2022-05-06
Removing files leftover from the project template

TylerMatteo:
### Summary
This PR just deletes some boilerplate components leftover from the project template

PR #160, opened 2022-05-09
Update jest config to report coverage and add some unit tests

TylerMatteo:
### Summary
This PR updates the jest config to report coverage for the unit tests. Once we get up to an acceptable coverage score, we should add a coverage threshold guard on future PRs. I also added tests for a handful of files that I worked on. I'll have more PRs like this as I find time to pay down our unit test tech debt but wanted to put a lid on this for now so that other can point out if I'm doing anything silly in these tests before I go too far :). 

Note that this PR also includes changes found in this PR - https://github.com/NYCPlanning/equity-tool/pull/159 - mainly deleting some placeholder and boilerplate files, so we'll want to merge that PR first.

**There is one small change in this PR to actual application code** - I made a small change to GeographyInfo so that it only attempts to get NTA info if the selected Geography is `NTA`. This was originally made to avoid having to do some unnecessary mocking my tests but I think it's a good change on it's own and saves us executing more code than is necessary.

### DIFF
--- a/src/components/DataPointCell/DataPointCell.spec.tsx
+++ b/src/components/DataPointCell/DataPointCell.spec.tsx
@@ -1,0 +1,79 @@
+import { ReactElement } from "react";
+import { render, screen, RenderOptions } from "@testing-library/react";
+import { Table, Tr, Tbody } from "@chakra-ui/react";
+import { DataPointCell } from "./DataPointCell";
+import { DataPoint } from "@schemas/dataPoint";
+
+interface TableWrapperProps {
+  children: ReactElement;
+}
+
+const TableWrapper = ({ children }: TableWrapperProps) => (
+  <Table>
+    <Tbody>
+      <Tr>{children}</Tr>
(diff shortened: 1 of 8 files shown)


godfreyyeung (approved):
Looks okay but I think @dhochbaum-dcp should take a look, too.

src/components/DataPointCell/DataPointCell.spec.tsx
```diff
@@ -0,0 +1,79 @@
...
+    <Tbody>
+      <Tr>{children}</Tr>
+    </Tbody>
+  </Table>
+);
+
+// Chakra Td elements must be rendered inside a Table by design. This simple wrapper around render()
+// keeps our tests DRY - see https://testing-library.com/docs/react-testing-library/setup/#custom-render
+// Might want to extract this to a common test utils folder so it can be used for other table components
+const renderInTable = (
```
godfreyyeung, line 22: kewl!

PR #168, opened 2022-07-06
9366 – Reset Community Data default boundary

bmarchena:
### Summary

Clicking on a boundary (eg. Community District, Borough, Citywide) in the Community Data map, navigating to the DRM, and then navigating back would cause the Community Data map to default to that boundary.

By setting the `LastCommunityDataGeography` state to null when clicking into the DRM map, we can preserve the default view and intended behavior.

Video preview:
https://user-images.githubusercontent.com/43344288/177577349-7d8b6fb5-1f32-4e1e-b75a-ab06097703d3.mov

#### Tasks/Bug Numbers
 - Fixes AB[#9366](https://dev.azure.com/NYCPlanning/ITD/_sprints/taskboard/Open%20Source%20Engineering/ITD/2022/Q3/Sprint%20N?workitem=9366)

### DIFF
--- a/src/pages/map/[view]/[geography].tsx
+++ b/src/pages/map/[view]/[geography].tsx
@@ -111,7 +111,7 @@ const MapPage = ({ initialRouteParams }: MapPageProps) => {
   const [lastBoroughGeoid, setLastBoroughGeoid] = useState<string | null>(null);
 
   const onDrmClick = () => {
-    setLastCommunityDataGeography(geography);
+    setLastCommunityDataGeography(null);
     setLastCommunityDataGeoid(geoid);
 
     ReactGA.event({


godfreyyeung (changes requested):
More clarity is needed from the Azure Bug writeup. 

The bug writeup indicates that the issue is clicking Community Data, while already on Community Data, will not switch the geography back to Community District. However the expected behavior of the app is the user shouldn't be able to click Community Data when that view is already active. 

We need more

src/pages/map/[view]/[geography].tsx
```diff
@@ -111,7 +111,7 @@ const MapPage = ({ initialRouteParams }: MapPageProps) => {
   const [lastBoroughGeoid, setLastBoroughGeoid] = useState<string | null>(null);
 
   const onDrmClick = () => {
-    setLastCommunityDataGeography(geography);
+    setLastCommunityDataGeography(null);
```
Comment (godfreyyeung) [inline src/pages/map/[view]/[geography].tsx:114]: setting this to null will break the feature for "remembering" last selected community data geography.

Comment (bmarchena) [reply src/pages/map/[view]/[geography].tsx:114]: i may have misunderstood the ticket but i think that feature is actually the exact bug being described so I'm a bit confused about what the intended behavior should be now

Comment (bmarchena) [reply src/pages/map/[view]/[geography].tsx:114]: Leo said that the behavior we have now is fine so I'll just close this out

PR #170, opened 2022-07-07
 9426 'Community Data' and 'DRM' are transparent when selected on mobile view

horatiorosa:
### Summary
A bug fix for a desktop navigation error was applied unnecessarily to the mobile experience causing unexpected styling in the mobile buttons..

#### Tasks/Bug Numbers
 - Fixes [AB#9426](https://dev.azure.com/NYCPlanning/ITD/_workitems/edit/9426)

#### BEFORE
error

#### AFTER
Screen Shot 2022-07-07 at 12 10 55 PM

### Technical Explanation
removing the disabled attribute logic from the mobile buttons in `src/components/Map/ViewToggle.tsx`, lines 46 and 64
- the logic was originally put in to prevent navigation errors found on the desktop when click between comm data and DRM per [AB#8003](https://dev.azure.com/NYCPlanning/ITD/_workitems/edit/8003) and [AB#8005](https://dev.azure.com/NYCPlanning/ITD/_workitems/edit/8005). The same conditions do not apply to the mobile navigation experience
 -  additionally, the disable attribute was overriding the button styles (the transparency)

https://user-images.githubusercontent.com/11340947/178515395-3289e98a-aa76-4e8a-a504-aa9f510325bd.mov

### DIFF
--- a/src/components/Map/ViewToggle.tsx
+++ b/src/components/Map/ViewToggle.tsx
@@ -31,7 +31,6 @@ export const ViewToggle = ({
         >
           <Button
             flex="1"
-            variant="toggle"
             backgroundColor="gray.50"
             height="100%"
             borderRadius=".5rem 0 0 0"
@@ -43,14 +42,15 @@ export const ViewToggle = ({
             onClick={onCommunityDataClick}
             isActive={view === "data"}
             isDisabled={view === "data"}
+            _hover={{ _disabled: { bg: "teal.50" } }}
             isFullWidth
             data-cy="communityDataBtn-mobile"
+            variant="mobileButton"
           >
             Community Data
           </Button>
           <Button
             flex="1"
-            variant="toggle"
             backgroundColor="gray.50"
             height="100%"
             borderRadius="0 .5rem 0 0"
@@ -62,8 +62,10 @@ export const ViewToggle = ({
             onClick={onDrmClick}
             isActive={view === "drm"}
             isDisabled={view === "drm"}
+            _hover={{ _disabled: { bg: "teal.50" } }}
             isFullWidth
             data-cy="drmBtn-mobile"
+            variant="mobileButton"
           >
             Displacement Risk Map
           </Button>
--- a/src/theme.tsx
+++ b/src/theme.tsx
@@ -74,7 +74,6 @@ const theme = extendTheme({
       baseStyle: {
         backgroundColor: "white",
         color: "gray.600",
-
         _hover: {
           backgroundColor: "teal.50",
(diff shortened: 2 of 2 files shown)


horatiorosa:
> Not sure this is the best approach. Functionality wise I think we still want users to not be able to redundantly tap the currently active view button. If there is a styling issue, we want to resolve that through styling and css .

@godfreyyeung 

A valid concern, I did initially investigate the CSS/Styling route to no avail. In ` src/components/Map/ViewToggle.tsx` the button referenced on line 31, I added the following:
```
            _disabled={{
              background: "teal.50",
              opacity: "1 !important"
            }}

```
to no effect. The disabled attribute adds `opacity: 0.4` which I cannot seem to override. As a spaghetti too, I changed the order of `isActive` to come after `isDisabled`. I didn't think it would work, but that's the place I am currently in 😬 😬

I am open to suggestions.

horatiorosa:
> That's interesting that it would only slap a .4 opacity for disabled. Have you looked into attributes like `bg` (adding a white background), or tried targeting it through `sx` or by slapping your own classes onto it? https://chakra-ui.com/docs/styled-system/the-sx-prop can you share a screenshot of the computed classes you see in chrome devtools?

I read the docs and `sx` with negative results.

```
sx={{
  "button:disabled": {
      opacity: "1 !important"
    },
}}
```
Even worse, digging into the issue further reveals that the DRM button disabled sets the background to initial (which I am guessing is white).

2 buttons, 2 sets of problems.

godfreyyeung (approved):
I like this css approach of using the a theme variant much better. Nice job

TylerMatteo:
I see that this is already merged but let's try to revisit this in the near future to get rid of the `!important`s. If we need those, it's probably a good sign that there is a bigger refactoring needed here.

horatiorosa:
@TylerMatteo  and @godfreyyeung 
I am not of fan of `!important` either but `isDisabled={view === "drm"}` adds an inline `disabled` attribute which necessitated this sort of nuclear option to override the disabled styles.

TylerMatteo:
> @TylerMatteo and @godfreyyeung I am not of fan of `!important` either but `isDisabled={view === "drm"}` adds an inline `disabled` attribute which necessitated this sort of nuclear option to override the disabled styles.

Yeah I hear you. I think the solution here would take a little more refactoring - IMO you shouldn't need an `isDisabled` at all, seeing as how these two buttons are effectively radio buttons because one and exactly one of them should be "active" at any given time. We can visit how we might refactor this using different Chakra components or hooks in the future.

PR #173, opened 2022-07-21
9752 Add RER files to the data download

dhochbaum-dcp:
### Summary
All three files zipped together for each PUMA, and downloaded when the user selects the PDF download on the Data Tables page for a PUMA. There will not be any changes made to downloads from the Data Tables page for Boro and Citywide, nor will there be changes to the download on the DRM.

image
image

#### Tasks/Bug Numbers
 - Completes [AB#9752](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/9752)

### Technical Explanation
PDFs are only available for districts, so added logic to check the geotype and serve up the appropriate text.


godfreyyeung (approved):
Looks okay but the title and description seem misleading..

PR #177, opened 2022-09-23
10606 – Display tooltips when hovering over geographies

bmarchena:
Resolves AB[#10606](https://dev.azure.com/NYCPlanning/ITD/_sprints/taskboard/Open%20Source%20Engineering/ITD/2022/Q3/Sprint%20T?workitem=10606).

This PR adds a tooltip when hovering over geographies on the map in both CD and DRM views. It adds an additional state variable to [geography].tsx which tracks the information about the currently hovered geography.

~~TODO:~~
~~- Currently there is an outstanding issue where the map rerenders when hovering over a geography in a view for the first time.~~
~~- The tooltip also does not currently disappear when hovering over an invalid geography.~~

https://user-images.githubusercontent.com/43344288/192009249-74a19178-5c8d-498d-bd58-8143fa549c19.mov


src/hooks/useLayers/useLayers.ts
```diff
@@ -191,6 +197,9 @@ export const useLayers = (): CartoLayer<any, any>[] | null => {
           toggleGeoSelect(id);
         }
       },
+      onHover: (info: any) => {
+        info.layer?.props.visible ? setTooltip(info) : setTooltip({});
```
TylerMatteo: Are you sure you need these `props.visible` checks? I think Deck only calls onHover for visible layers by default

src/components/Map/Map.tsx
```diff
@@ -70,6 +91,31 @@ export const Map = ({ layers, parent }: MapProps) => {
         mapboxApiAccessToken={process.env.NEXT_PUBLIC_MAPBOX_TOKEN}
         mapStyle={baseMap}
       ></ReactMapGL>
+      {tooltipText && !tooltipText.includes("etc") && (
+        <div
+          style={{
+            position: "absolute",
```
TylerMatteo, line 102: It looks like the tooltip should be centered under the cursor in the Figma. It also has the little [arrow](https://chakra-ui.com/docs/components/tooltip/usage#with-arrow) that Chakra's tooltip supports but Design might be cool with dropping that if we are having trouble using Chakra's tooltip component within the map.

src/components/Map/Map.tsx
```diff
@@ -9,18 +9,24 @@ import baseMap from "@data/basemap.json";
...
   apiKey: process.env.NEXT_PUBLIC_CARTO_API_KEY,
 });
 
-type MapProps = Pick<DeckGLProps, "layers" | "parent">;
+type DeckProps = Pick<DeckGLProps, "layers" | "parent">;
 
-export const Map = ({ layers, parent }: MapProps) => {
-  const view = useView();
+interface MapProps extends DeckProps {
+  hoverInfo: any | null;
```
TylerMatteo, line 24: Maybe we can replace this `any` with a proper type as well? This one might be a littler trickier.

TylerMatteo (changes requested):

src/hooks/useLayers/useLayers.ts
bmarchena: good call, thought i needed them to clear the tooltip state but it's not necessary

src/components/Map/Map.tsx
```diff
@@ -39,6 +45,21 @@ export const Map = ({ layers, parent }: MapProps) => {
         bearing: 0,
       };
 
+  let tooltipText = usePumaInfo(
+    hoverInfo?.object?.properties.puma
+  )?.neighborhoods;
+
+  switch (geography) {
+    case "borough":
+      tooltipText = hoverInfo?.object?.properties.boroname;
```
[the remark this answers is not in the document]
bmarchena: yup just saw the copy and updated in my last commit

src/components/Map/Map.tsx
bmarchena: gonna try using the Chakra tooltip but without a built-in component to handle that style I could see it being a non-trivial lift to get the arrow centered. digging into it

bmarchena: looks like the Chakra Tooltip component only really works when it's wrapping the content we want to display a tooltip over and it isn't really gonna work with our implementation of the EDDE map/layers. maybe we can punt on the arrow for now if that's ok with design?

TylerMatteo: Understood. I'll ask design about dropping the arrow but I doubt it will be a problem. Were you able to get the tooltip to be centered under the mouse without the arrow? I could see how that would be tricky without having Chakra's props to lean on so that might also be negotiable.

bmarchena: just added that in the last commit, had to use add a ref but it's working well now

TylerMatteo:
@bmarchena  this all looks good to me but I'm going to hold off on approving this until I merge in a branch I'm working on to fix the builds. You may have noticed that the GH actions that kick off when you push to your branch are failing to actually do preview deploys. Once that's in, I can help you rebase if there are any conflicts.

src/hooks/usePumaInfo/usePumaInfo.ts
```diff
@@ -4,8 +4,9 @@ import { hasOwnProperty } from "@helpers/hasOwnProperty";
 
 export type pumaInfo = typeof pumas["3701"];
 
-export const usePumaInfo = (): pumaInfo | null => {
-  const geoid = useGeoid();
+export const usePumaInfo = (id?: string | null): pumaInfo | null => {
+  let geoid = useGeoid();
+  if (id) geoid = id;
```
bmarchena, line 9: @TylerMatteo I actually ended up adding this back as an optional param if that's cool with you. The useGeoid hook will always return a null geoid on the default map page so we'll have to pass that along for each hovered geo for the tooltip to work as intended.

TylerMatteo: Ah, gotcha. Yeah this is fine. You don't _technically_ need a hook here so if we wanted to be really strict we could move some of this code to a generic "getPumaInfo" function that you could call directly and my hook would call but that's probably a bit over kill.

TylerMatteo (approved):

PR #178, opened 2022-09-27
Task/10651 geo labels on map

dhochbaum-dcp:
### Summary
As a user, I want to be able to view geography labels at relevant zoom levels so that I can more easily navigate the map, similarly to how I see labels on PFF.

For the Community Data map, this includes starting with the borough names, then the PUMA names at medium zoom levels, and adding in more specific NTA labels as I zoom in closer.

For the DRI map, this includes starting with the borough names, then the NTA labels as I zoom in closer

Acceptance Criteria:
1) at lower zoom levels, borough names display
2) at mid zoom levels, PUMA labels (names only) display
3) at higher zoom levels, NTA labels also appear in smaller font than PUMA

Figma for reference:

[https://www.figma.com/file/8ExFFiN7KDe6vzoQchnOJC/EDDE-1.1?node-id=226%3A23801](https://www.figma.com/file/8ExFFiN7KDe6vzoQchnOJC/EDDE-1.1?node-id=226:23801)

#### Tasks/Bug Numbers
 - Implements [AB#10651](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/10651)


src/hooks/useLayers/createCompositeLayers.ts
```diff
@@ -0,0 +1,127 @@
...
+        }
+      ),
+    ];
+  }
+
+  //https://deck.gl/docs/api-reference/core/composite-layer
+  filterSubLayer({ layer, viewport }: { layer: any; viewport: any }) {
+    if (layer.id.slice(-14) === "_pumatextlayer") {
+      return 10.5 < viewport.zoom && viewport.zoom < 13;
+    } else if (layer.id.slice(-13) === "_ntatextlayer") {
```
TylerMatteo, line 117: I wouldn't hold up the PR over this but maybe give some thought to how else you could identify the layer here. Having to slice up strings to do text matching can get bug-prone over time.

src/hooks/useLayers/useLayers.ts
```diff
@@ -81,10 +86,12 @@ export const useLayers = (): CartoLayer<any, any>[] | null => {
         }
       },
     }),
-    new CartoLayer({
+    new LabeledCartoLayer({
+      passedId: BOROUGH,
       visible: geography === BOROUGH,
       type: MAP_TYPES.QUERY,
-      id: BOROUGH,
+      // id: BOROUGH,
```
TylerMatteo, line 93: Just a heads up to remember to clean up any leftover comments before this gets final approval.

TylerMatteo (changes requested):

src/hooks/useLayers/createCompositeLayers.ts
```diff
@@ -0,0 +1,127 @@
...
+            sdf: true,
+          },
+          outlineWidth: 1,
+          outlineColor: [255, 255, 255, 255],
+          maxWidth: 800,
+          sizeUnits: "meters",
+        }
+      ),
+      new TextLayer(
+        {},
```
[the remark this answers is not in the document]
dhochbaum-dcp: I can't recall why those are there - I've removed them.

src/hooks/useLayers/useLayers.ts
dhochbaum-dcp: Removed.

src/hooks/useLayers/createCompositeLayers.ts
dhochbaum-dcp: These match the strings that are added in the creation of the layer above.  There really isn't anything else that exists that is unique that it makes sense to use (I could check maxWidth or something, but that is really a layer of abstraction that makes things more bug-prone).  I could add a field to each layer specifically for this, but I'm not sure that would make things any less complicated.

TylerMatteo (approved):

PR #185, opened 2022-10-07
Task/10658 data page sticky header

TylerMatteo:
### Summary
This PR adds the redesign of the "header" portion of the data tables page.
* Header is made "sticky" when user scrolls, with different portions of it being sticky depending on users screen size
* Moves data reliability message into a popover
* Adds new tab-like buttons for switching between subgroups, instead of the existing dropdown select
* Various other style and positioning tweaks

I also added a new `useGeoidDescription` hook to consolidate some of the logic for getting the human-readable "labels" for geoids into one place.


bmarchena (dismissed):
LGTM, nice idea calling the useGeoid hook in usePumaInfo instead of passing it as a param each time

PR #187, opened 2022-10-11
10653 – Add outline and cursor on geography hover

bmarchena:
### Summary
Adds highlighted outline around geographies on hover. Changes cursor when hovering over valid geographies.
~~- Cursor still needs to be added~~
~~- There is still an outstanding issue regarding performance.~~

#### Tasks/Bug Numbers
 - Fixes AB[#10653](https://dev.azure.com/NYCPlanning/ITD/_sprints/taskboard/Open%20Source%20Engineering/ITD/2022/Q4/Sprint%20U?workitem=10653)

https://user-images.githubusercontent.com/43344288/195644028-642b7c4b-e29a-434e-bfbb-f83821a0833a.mov


src/components/Map/Map.tsx
```diff
@@ -12,37 +12,21 @@ import { useWindowWidth } from "@react-hook/window-size";
...
-    };
-  };
-}
-
-interface MapProps extends DeckProps {
-  hoverInfo: HoverType | null;
-}
-
-export const Map = ({ layers, parent, hoverInfo }: MapProps) => {
+type DeckProps = DeckGLProps<"parent">;
```
TylerMatteo, line 23: This is a little nitpicky but can you change this back to being called `MapProps`? It's a common convention in TS React to name your props types like "NameOfComponentProps"

src/components/Map/Map.tsx
```diff
@@ -12,37 +12,21 @@ import { useWindowWidth } from "@react-hook/window-size";
...
+
+export const Map = ({ parent }: DeckProps) => {
+  const [hoverInfo, setHoverInfo] = useState<any | null>(null);
```
TylerMatteo, line 26: I'm fine with leaving this as `any` for now if you're having hard time typing it but just curious why this change? Could you use the `HoverType` interface that you got rid of here?

TylerMatteo (changes requested):
Just one change I'd like to make but otherwise, lgtm! I tested the preview deployment and the performance seems to be at least as good as production.

src/components/Map/Map.tsx
bmarchena: I tried that originally but got a funky type mismatch error because of the Dispatch action
Screen Shot 2022-10-17 at 11 13 41 AM

TylerMatteo: I see. All good. I'll take a crack at refactoring this after your merge.

TylerMatteo (approved):

PR #191, opened 2022-10-21
11091 contact page with form and new footer with links and newsletter sign-up

horatiorosa:
### Summary
Created: 
 -  a new contact form to elicit feedback and comments from users.
 - a new footer for easier navigation when users have scrolled closer to the bottom of the page.
 - a new newsletter sign-up for to enable users be informed of updates and changes to EDDE.
 - update nav bar to include new link to the contact page.

#### Contact page and empty form
Screen Shot 2022-11-03 at 11 07 50 AM

#### Contact page after form has been sent
Screen Shot 2022-11-03 at 11 06 38 AM

#### Tasks/Bug Numbers
 Fixes:
 -  [AB#11091](https://dev.azure.com/NYCPlanning/ITD/_workitems/edit/11091)
 -  [AB#11252](https://dev.azure.com/NYCPlanning/ITD/_workitems/edit/11252)


package.json
```diff
@@ -22,8 +22,6 @@
     "lint:tsc": "tsc --noEmit && tsc --noEmit --project ./cypress/tsconfig.json",
     "test": "NODE_ENV=test jest -c test/jest.config.js",
     "e2e:open": "cypress open",
-    "e2e:run": "cypress run --config video=false",
```
TylerMatteo, line 25: ???

horatiorosa: I put it back, sorry!

src/pages/api/feedback.ts
```diff
@@ -0,0 +1,41 @@
...
+type ResponseData = {
+  responseMessage: any;
+};
+
+export default async (
+  req: NextApiRequest,
+  res: NextApiResponse<ResponseData>
+) => {
+  try {
+    if (req.method === "POST") {
```
TylerMatteo, line 14: This is correct for enforcing `POST` but you need an else. Just sending an error with a 404 code for any other method should be fine.

src/schemas/feedbackSchema.ts
```diff
@@ -0,0 +1,20 @@
+import * as Yup from "yup";
+
+export const feedbackSchema = Yup.object({
+  to: Yup.string().default(""),
```
TylerMatteo, line 4: I think you can get rid of this, no? This schema is meant to be a representation of your form, which seems to only have 3 fields - name, email, and feedback.

src/schemas/feedbackSchema.ts
```diff
@@ -0,0 +1,20 @@
...
+    .required(),
+  email: Yup.string()
+    .default("")
+    .email("Please enter a valid email address.")
+    .required("This is a required field."),
+  feedback: Yup.string()
+    .default("")
+    .min(12, "Feedback should be at least 12 characters.")
+    .required(),
+  subject: Yup.string().default("EDDE Feedback"),
```
TylerMatteo, line 17: Same as `to`.

src/components/ContactForm/ContactForm.tsx
```diff
@@ -0,0 +1,137 @@
...
+    try {
+      await axios({
+        method: "post",
+        url: "/api/feedback",
+        data: feedbackBody,
+      }).then(function () {
+        setShowContactForm(!showContactForm);
+      });
+    } catch (error) {
+      console.error("submit error", error);
```
TylerMatteo, line 53: Thoughts on how we should properly handle this error state? We'll want to at least show an error to the user and possibly also reset the form's state so that they can try again.

src/pages/api/feedback.ts
```diff
@@ -0,0 +1,41 @@
+import type { NextApiRequest, NextApiResponse } from "next";
+import * as sgMail from "@sendgrid/mail";
+import { feedbackSchema } from "@schemas/feedbackSchema";
+
+type ResponseData = {
```
TylerMatteo, line 5: I think we can improve the variable naming and typing here. Naming is always subjective but I think "response" is a little redundant in this content. And whatever the name is, I think the single property of this type could be more specifically typed than `any`. `responseMessage` implies to me that it's just a string but I see below that you're passing in the whole response object that we get from SG.

src/pages/api/feedback.ts
```diff
@@ -0,0 +1,41 @@
...
+      const data: any = {
+        to: process.env.TO_EDDE_EMAIL,
+        from: process.env.FROM_EDDE_EMAIL,
+        subject: body.subject,
+        text: feedbackMessage.replace(/\n\r/g, `<br>`),
+      };
+
+      const responseMessage = await sgMail.send(data);
+
+      res.status(res.statusCode).json({ responseMessage });
```
TylerMatteo, line 35: We probably don't want to return the whole response that comes back from SG. I'm assuming they're giving us back some sort of object? You might want to return just the info we need on the front end.

src/pages/api/feedback.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+  } catch (error: any) {
+    console.error(error);
+    res.status(res.statusCode).json(error);
```
TylerMatteo, line 39: Sorta the same issue as in the try block - This is going to send back the whole response object (headers and all). A couple approaches you could take:
* You might want to return just a generic error message with only as much detail as you need on the FE
* Only return party of the error object. [SG's docs](https://github.com/sendgrid/sendgrid-nodejs/tree/main/packages/mail#quick-start-hello-email) seem to suggest there would be a `body` property at `error.response.body`

TylerMatteo (changes requested):

src/schemas/feedbackSchema.ts
horatiorosa: Yea, I think we can cut the TO from this schema.

src/components/ContactForm/ContactForm.tsx
horatiorosa: Got a quick mock-up from design and it's implemented.

src/pages/api/feedback.ts
horatiorosa: Not sure how to fix this one. If I set the type as a Object, I get errors here if I try `responseMessage.body}`

src/schemas/feedbackSchema.ts
TylerMatteo: You can have that without have that here, no? Remember - this schema describes the actual form you're showing to the user. If there is not an input box on the screen asking the user for something called "subject", you should not have a property here. If you want to always set this string as the subject of the email, that can be done directly in the API handler.

src/components/Contact/ContactForm.tsx
```diff
@@ -0,0 +1,142 @@
...
+    try {
+      await axios({
+        method: "post",
+        url: "/api/feedback",
+        data: feedbackBody,
+      }).then(function () {
+        setShowContactForm(!showContactForm);
+      });
+    } catch (error) {
+      console.error("submit error", error);
```
TylerMatteo, line 50: You can get rid of this now. Logging things in production is pretty much never a good habit to get into. At best, it's spamming users browsers, at worst, it exposes info we don't want to expose

src/components/Contact/ContactForm.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  const {
+    register,
+    handleSubmit,
+    formState: { errors, isSubmitting },
+  } = useForm<Feedback>({
+    resolver: yupResolver(feedbackSchema),
+  });
+
+  const characterCount = (e: any) => {
```
TylerMatteo, line 33: Won't hold up merging for this but you should be able to type the `e`. It would be some sort of HTML Event object. Also, this is a little subjective but think about naming event handler functions in a way that suggests that they're handlers...like`onFeedbackChange` or something. I just think it makes things clearer because at first glance one might assume that `characterCount` is a number, not a function.

TylerMatteo (dismissed):

src/components/Contact/ContactForm.tsx
horatiorosa: good point. I changed the func. name to `getCharacterCount`.

TylerMatteo (approved):

PR #195, opened 2022-11-01
10993a Improvements to burger menu on mobile

dhochbaum-dcp:
### Summary
This fix adds an underline to "Home" when the user is on the map page.

#### Tasks/Bug Numbers
 - Fixes [AB#10993](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/10993)

### DIFF
--- a/src/components/Header/NavLink/NavLink.tsx
+++ b/src/components/Header/NavLink/NavLink.tsx
@@ -14,6 +14,10 @@ export const NavLink = ({
   ...buttonProps
 }: NavLinkProps): JSX.Element => {
   const { pathname } = useRouter();
+  const isHomeLinkOnMapPage =
+    pathname === "/map/[view]/[geography]" && href.slice(0, 5) === "/map/"
+      ? true
+      : false;
   return (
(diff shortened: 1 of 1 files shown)


src/components/Header/NavLink/NavLink.tsx
```diff
@@ -14,6 +14,10 @@ export const NavLink = ({
   ...buttonProps
 }: NavLinkProps): JSX.Element => {
   const { pathname } = useRouter();
+  const isHomeLinkOnMapPage =
+    pathname === "/map/[view]/[geography]" && href.slice(0, 5) === "/map/"
```
TylerMatteo, line 18: Not something you need to change but just fyi, if the latter part of your ternary is just `? true : false`, you likely don't need a ternary at all. Simply `pathname === "/map/[view]/[geography]" && href.slice(0, 5) === "/map/"` would be functionally the same here I think.

TylerMatteo (approved):

PR #197, opened 2022-11-03
11246 Main Navigation Items Link to Top of Pages

dhochbaum-dcp:
### Summary
As a user, I want main navigation items to link to the top of the pages, so that I can orient myself. 
 - For following navigation items - "About," "Methods & Sources," and "Contact," - link to the top of each page, rather than remembering the user's scroll position
 - "Home" can preserve scroll position if that maintains the user's map selection

#### Tasks/Bug Numbers
 - Completes [AB#11246](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/11246)

### Technical Explanation
~~Refactored the components from functions to classes in order to make use of component lifecycle, specifically componentDidMount.  The function in componentDidMount brings the page back to the top the same way that the back to top button does.~~
useEffect hook brings pages back to top.

### DIFF
--- a/src/pages/about.tsx
+++ b/src/pages/about.tsx
@@ -1,3 +1,4 @@
+import { useEffect } from "react";
 import {
   Flex,
   Box,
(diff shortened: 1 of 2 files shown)


horatiorosa:
I was wondering if you saw this possible solution for scrolling to the top on page changes:
https://github.com/vercel/next.js/discussions/15337

src/pages/about.tsx
```diff
@@ -11,269 +11,287 @@ import { ExternalLinkIcon } from "@chakra-ui/icons";
...
-        <Feedback
-          flex={1}
-          py={10}
-          fontSize="1rem"
-          display={{ base: "none", md: "block", xl: "block" }}
-        />
-      </Flex>
-    </Center>
+class AboutPage extends React.Component {
+  constructor(props: any) {
```
godfreyyeung, line 17: Is this necessary? Would(n't) it automatically call super by default if you didn't define a constructor here?

src/pages/methods.tsx
```diff
@@ -12,480 +12,499 @@ import {
...
-                    label: "NYC Population: Geographic Reference",
-                  });
-                }}
-              >
-                NYC Population: Geographic Reference
-              </Link>
-              .
-            </Text>
+class MethodsPage extends React.Component {
+  constructor(props: any) {
```
godfreyyeung, line 18: same comment here

src/pages/about.tsx
dhochbaum-dcp: Looks like you're right, I've updated this to remove it.

dhochbaum-dcp:
> I was wondering if you saw this possible solution for scrolling to the top on page changes: [vercel/next.js#15337](https://github.com/vercel/next.js/discussions/15337)

I did not, but it appears to be a much more complicated solution, and not necessarily effective.

src/pages/about.tsx
```diff
@@ -11,269 +11,284 @@ import { ExternalLinkIcon } from "@chakra-ui/icons";
...
-        </Box>
-        <Feedback
-          flex={1}
-          py={10}
-          fontSize="1rem"
-          display={{ base: "none", md: "block", xl: "block" }}
-        />
-      </Flex>
-    </Center>
+class AboutPage extends React.Component {
```
TylerMatteo, line 16: I'm going to pull this and see if I can get it the same behavior working with `useEffect`. Reeeaalllly don't want to introduce class components if we can help it.

TylerMatteo (changes requested):

TylerMatteo (approved):
Lgtm. I think needing a "back to top" element just to have something to scroll is less than ideal but it's a good enough workaround for now, especially considering that this is apparently a [known issue](https://github.com/vercel/next.js/issues/28778) in Next at the moment.

PR #207, opened 2023-02-14
Update data dictionary

TangoYankee:
### Summary

**Issue context**
- Link to data dictionary is outdated
- Multiple instances of the link exist

**Proposed Fix**
- Create a constant to hold the link
  - This makes it easier to identify all the places where the link is used
  - It also makes it faster to update the link everywhere
- Replace the outdated link with the current link   

#### Tasks/Bug Numbers
 - Fixes [AB#12256](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/12256)

### Technical Explanation
- The constants folder appeared to be the best place to hold the links
- I used an object, to allow for more external links to be added later

### Any other info you think would help a reviewer understand this PR?
- I realize this might be overkill. So, I also have a version where I just swap the old link for the new one with a 'find and replace'. I can submit that one for review, if it's preferable.

### DIFF
--- a/src/constants/externalLink.ts
+++ b/src/constants/externalLink.ts
@@ -1,4 +1,2 @@
-export const EXTERNAL_LINK = {
-  DATA_DICTIONARY:
-    "https://www.nyc.gov/assets/planning/download/pdf/data-maps/edde/edde-data-dictionary.pdf",
-};
+export const DATA_DICTIONARY =
+  "https://www.nyc.gov/assets/planning/download/pdf/data-maps/edde/edde-data-dictionary.pdf";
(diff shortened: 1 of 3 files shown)


src/constants/externalLink.ts
```diff
@@ -0,0 +1,4 @@
+export const EXTERNAL_LINK = {
```
TylerMatteo, line 1: Overall, I'm cool with putting this in a constant since it shows up in a couple spots.

This is very nitpicky and subjective but could you have this file just export strings? So this would become something like `export const DATA_DICTIONARY = "https://www.nyc.gov/assets/planning/download/pdf/data-maps/edde/edde-data-dictionary.pdf"`, and the import would change accordingly. Other links that are added later can just be new named exports in this same file.

TylerMatteo (changes requested):

src/constants/externalLink.ts
TangoYankee: change is in

horatiorosa (commented):
🚀 
this is better than a swap and replace, much easier to update in the future, if it changes again.

PR #209, opened 2023-02-17
Bug/12305 data table header heights

TangoYankee:
### Summary
Only table body/data rows were being standardized across all vintages. The table header rows also needed to be standardized.

This solution assumes the content and structure of the first vintage will be appropriate for all subsequent vintages. This assumption holds true throughout the current site layout and data content. This assumption may need to be revisited if there are significant changes to the data or layout. 

It also adds a note on where to find the development credentials

#### Tasks/Bug Numbers
 - Fixes [AB#12305](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/12305)

### Technical Explanation
This takes the same approach used to standardize the data row heights. It collects the heights of the first vintage and applies it to the rest of the vintages. If am I reading the code correctly, it deviates from the original approach in that it does not set the height for the first vintage. Instead, it leaves the height of the first vintage undefined in order to let it remain responsive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ npm i
 This command should install all necessary dependencies and set up Husky to run on all commits
 
 2. Create `.env.local` and set the required environment variables.
+   - Local development credentials are available within the NYC planning password vault under "EDDE Credentials"
 
 
 ### Running locally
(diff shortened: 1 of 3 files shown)


TangoYankee:
Marking as draft; depending on how #AB/12307 is solved, these changes may no longer apply

PR #210, opened 2023-02-23
Bug/12307 sticky data column bad glue

TangoYankee:
### Summary
Only table body/data rows were being standardized across all vintages. The table header rows also needed to be standardized.

This solution assumes the content and structure of the first vintage will be appropriate for all subsequent vintages. This assumption holds true throughout the current site layout and data content. This assumption may need to be revisited if there are significant changes to the data or layout. 

It also adds a note on where to find the development credentials

#### Tasks/Bug Numbers
- Fixes AB#12307
 - Fixes [AB#12305](https://dev.azure.com/NYCPlanning/cc280b0d-40a0-4689-b852-2e6247f1af50/_workitems/edit/12305)

### Technical Explanation
This takes the same approach used to standardize the data row heights. It collects the heights of the first vintage and applies it to the rest of the vintages. If am I reading the code correctly, it deviates from the original approach in that it does not set the height for the first vintage. Instead, it leaves the height of the first vintage undefined in order to let it remain responsive.


src/components/VintageList/GrayCard.tsx
```diff
@@ -0,0 +1,16 @@
+import { Box } from "@chakra-ui/react";
+
+const GrayCard = () => (
+  <Box
+    backgroundColor="gray.50"
+    position="absolute"
+    height="100%"
+    minWidth={{ base: "calc((100vw - 26px) / 3)", md: "13.5rem" }}
+    maxWidth={{ base: "calc((100vw - 26px) / 3)", md: "13.5rem" }}
+    display={{ base: "none", md: "block" }}
```
TangoYankee, line 10: Rather than using `display: none` for small screens, it might be better to wrap the whole `Box` component in a  `<Show above=''md'>` [component](https://chakra-ui.com/docs/components/show-hide)

src/components/VintageList/VintageList.tsx
```diff
@@ -16,42 +30,75 @@ export const VintageList = ({
...
-      heights.push(node.clientHeight);
-    });
-    setRowHeights(heights);
+    setBodyRowHeights(getBodyClientHeights(labelRef));
   }, [category, subgroup]);
 
+  // Update row heights whenever category or subgroup changes, or reliability is toggled
+  useEffect(() => {
+    const heights = getHeaderClientHeights(firstTableRef);
+    setHeaderRowHeights(heights);
```
TangoYankee, line 60: creating a 'heights' variable isn't necessarily helpful. (This is a leftover from a previous solution strategy). It might be cleaner to just pass the call to `getHeaderClientHeights` directly into `setHeaderRowHeight`

src/components/VintageLabels/VintageLabels.tsx
```diff
@@ -0,0 +1,69 @@
...
+        >
+          <Thead opacity={0}>
+            <Tr>
+              <Th
+                rowSpan={1}
+                colSpan={2}
+                display="table-cell"
+                minWidth={"13.5rem"}
+                maxWidth={"13.5rem"}
+                height={`${headerHeight}px`}
```
TangoYankee, line 41: From the [layout-width-and-height docs](https://chakra-ui.com/docs/styled-system/style-props#layout-width-and-height), specifying px shouldn't be necessary- it should default to `px`

PR #211, opened 2023-03-02
Bug/12307 two tables

TangoYankee:
#### Tasks/Bug Numbers
 -


TylerMatteo (approved):
lgtm. Thanks for sticking with this one. This is definitely easier to follow than the hold "responsive" tables.

PR #223, opened 2023-05-11
Comment out e2e GH action step and set environment URLs for staging and production

TylerMatteo:
### Summary
This PR does two things:
* Comments out the `e2e` job in each action workflow. These tests have been failing for a while and until we figure out a way to have them behave correctly inside of GH action runners, I think it's best to comment out the step so that the entire workflow run isn't classified as "failed". I [looked into it](https://github.com/actions/runner/issues/2347), but there doesn't seem to be a way to "allow" a particular job in a workflow to fail.
* Sets a `url` under `environment` in the pr_merged and release_and_deploy workflows. This should trigger GH to associate that URL with the environments listed on the homepage of the repo. Note that I didn't do that the feature_branch workflow because that will a little more involved, as those urls are dynamic but we should do that in the near future as well. Here's some [documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-output-as-url) and a [SO post](https://stackoverflow.com/a/67385569) that should help with that.

### DIFF
--- a/.github/workflows/feature_branch.yml
+++ b/.github/workflows/feature_branch.yml
@@ -37,37 +37,37 @@ jobs:
       - name: Run tests
         run: npm run test
 
-  e2e:
-    name: 🧪 End To End Tests
-    needs:
-      - lint
-    runs-on: ubuntu-latest
-    environment: development
-    env:
-      NETLIFY_SITE_ID: ${{ secrets.NETLIFY_SITE_ID }}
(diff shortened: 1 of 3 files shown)


TangoYankee:
For the sake of due diligence, I'd like to verify a development build will pass with these changes. Would you be open to renaming the branch to follow the `task/[0-9]+-*` pattern?

